PR #15, opened 2017-01-04
more itertools defined

AlexEKoren:
Added cycle, dropwhile, from_iterable, ifilter, ifilterfalse, and takewhile in itertools


AlexEKoren:
I signed it!

ns-cweber:
Might be good to have some test cases as well.

AlexEKoren:
Done! Thanks for the feedback

ns-cweber:
Looks like you've committed some `.pyc` files by accident?

lib/itertools_test.py
```diff
@@ -16,6 +16,70 @@
...
+  limit = 10 * len(arg)
+  counter = 0
+  for x in itertools.cycle((0, 1, 2)):
+    got.append(x)
+    counter += 1
+    if counter == limit:
+      break
+  assert tuple(got) == want, 'tuple(cycle%s) == %s, want %s' % (arg, tuple(got), want)
+
+def testDropwhile():
```
ns-cweber: `def TestDropwhile`?

AlexEKoren:
Good catch. Should be fixed. Changed the .gitignore as well.

ns-cweber:
Good work. This all looks good to me, save for the aforementioned typo.

AlexEKoren:
Awesome! I signed the CLA, but not sure why it's not going through. What can I do rectify?

trotterdylan:
Really appreciate you taking this on!

Not sure what's up with the CLA issue, but did you see this comment? https://github.com/google/grumpy/pull/15#issuecomment-270507818

Do you have multiple emails associated with your GitHub account? If so, then make sure that the email that's associated with the git commits is the same one on the CLA?

AlexEKoren:
Yep, I double checked. Same e-mail (capitalization is different, but Github won't let me change it. Guessing that's not the issue)

trotterdylan:
Strangely, it looks like one of your commits is not like the other: 

9sarmkau43u

The one by AlexEKoren is fine, but the others aren't?

AlexEKoren:
@trotterdylan my CLI was setup with another e-mail account. Rewrote old commits with correct credentials! Wow, that was a weird one, great catch.

trotterdylan:
Awesome. This all looks great. Can you make sure that "make precommit" is clean?

AlexEKoren:
Just to make sure I understand correctly: Run 'make precommit' in the root directory and make sure it runs without error?

AlexEKoren:
I get the following output. Thoughts? Tried to make clean as well.

build/src/grumpy/builtin_types.go:587: z.Text undefined (type big.Int has no field or method Text)
build/src/grumpy/long.go:131: x.Text undefined (type *big.Int has no field or method Text)
build/src/grumpy/long.go:281: toLongUnsafe(o).value.Text undefined (type big.Int has no field or method Text)
build/src/grumpy/long.go:289: toLongUnsafe(o).value.Text undefined (type big.Int has no field or method Text)
make: *** [build/pkg/darwin_amd64/grumpy.a] Error 1

trotterdylan:
That's not a problem with your code. It may be that you have an older Go toolchain. Can you paste the output of "go version"? Mine is:

$ go version
go version go1.7.4 darwin/amd64

AlexEKoren:
Oops! Was a while since I updated Go. 

'make precommit' runs with exit status 0!

trotterdylan:
Awesome! Thanks for working on this!

Issue #22, opened 2017-01-04
from __future__ import print_function


paulsmith:
None of the future features are implemented, including the print function.

trotterdylan:
https://github.com/google/grumpy/pull/23 has been merged. Thanks again for implementing this!

PR #23, opened 2017-01-04
Implement print_function future feature

paulsmith:
This allows the import statement `from __future__ import print_function` to work, by setting a flag on the parser to treat print statement/keyword as a syntax error, and adding support to the Go runtime as a builtin.

Addresses #22

### DIFF
--- a/compiler/stmt.py
+++ b/compiler/stmt.py
@@ -56,9 +56,8 @@ redundant_future_features = ["generators", "with_statement", "nested_scopes"]
 late_future = 'from __future__ imports must occur at the beginning of the file'
 
 
-# import_from_future processes a future import statement, returning the set of
-# flags it defines.
 def import_from_future(node):
+  """processes a future import statement, returning set of flags it defines."""
   assert isinstance(node, ast.ImportFrom)
   assert node.module == '__future__'
   flags = 0
@@ -113,7 +112,7 @@ def visit_future(node):
         done = True
     elif isinstance(node, ast.Expr) and not found_docstring:
       e = node.value
-      if not isinstance(e, ast.Str):
+      if not isinstance(e, ast.Str): # pylint: disable=simplifiable-if-statement
         done = True
       else:
         found_docstring = True
--- a/compiler/stmt.py
+++ b/compiler/stmt.py
@@ -32,6 +32,28 @@ _NATIVE_TYPE_PREFIX = 'type_'
 _nil_expr = expr.nil_expr
 
 
+# Parser flags, set on 'from __future__ import *', see parser_flags on
+# StatementVisitor below. Note these have the same values as CPython.
+FUTURE_DIVISION         =  0x2000
+FUTURE_ABSOLUTE_IMPORT  =  0x4000
+FUTURE_PRINT_FUNCTION   = 0x10000
+FUTURE_UNICODE_LITERALS = 0x20000
+
+# Names for future features in 'from __future__ import *'. Map from name in the
+# import statement to a tuple of the flag for parser, and whether we've (grumpy)
+# implemented the feature yet.
+future_features = {
+  "division"         : (FUTURE_DIVISION, False),
+  "absolute_import"  : (FUTURE_ABSOLUTE_IMPORT, False),
+  "print_function"   : (FUTURE_PRINT_FUNCTION, True),
+  "unicode_literals" : (FUTURE_UNICODE_LITERALS, False),
+}
+
+# These future features are already in the language proper as of 2.6, so
+# importing them via __future__ has no effect.
+redundant_future_features = ["generators", "with_statement", "nested_scopes"]
+
+
 class StatementVisitor(ast.NodeVisitor):
   """Outputs Go statements to a Writer for the given Python nodes."""
 
@@ -41,6 +63,7 @@ class StatementVisitor(ast.NodeVisitor):
     self.block = block_
     self.writer = util.Writer()
     self.expr_visitor = expr_visitor.ExprVisitor(self.block, self.writer)
+    self.parser_flags = 0
 
   def generic_visit(self, node):
     msg = 'node not yet implemented: {}'.format(type(node).__name__)
@@ -286,6 +309,21 @@ class StatementVisitor(ast.NodeVisitor):
                 mod.expr, self.block.intern(name))
             self.block.bind_var(
                 self.writer, alias.asname or alias.name, member.expr)
+    elif node.module == '__future__':
+      for alias in node.names:
+        name = alias.name
+        if name in future_features:
+          (flag, implemented) = future_features[name]
+          if not implemented:
+            msg = 'future feature {} not yet implemented by grumpy'.format(name)
+            raise util.ParseError(msg)
+          self.parser_flags |= flag
+        elif name in redundant_future_features:
+          # no-op
+          pass
+        else:
+          msg = 'future feature {} is not defined'.format(name)
+          raise util.ParseError(node, msg)
     else:
       # NOTE: Assume that the names being imported are all modules within a
       # package. E.g. "from a.b import c" is importing the module c from package
@@ -305,6 +343,9 @@ class StatementVisitor(ast.NodeVisitor):
     self._write_py_context(node.lineno)
 
   def visit_Print(self, node):
+    if self.parser_flags & FUTURE_PRINT_FUNCTION:
+      raise util.ParseError(
+        node, 'syntax error (print is not a keyword)')
     self._write_py_context(node.lineno)
     with self.block.alloc_temp('[]*πg.Object') as args:
       self.writer.write('{} = make([]*πg.Object, {})'.format(
--- a/compiler/stmt.py
+++ b/compiler/stmt.py
@@ -57,7 +57,7 @@ late_future = 'from __future__ imports must occur at the beginning of the file'
 
 
 def import_from_future(node):
-  """processes a future import statement, returning set of flags it defines."""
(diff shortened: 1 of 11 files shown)


compiler/stmt.py
```diff
@@ -286,6 +309,21 @@ def visit_ImportFrom(self, node):
                 mod.expr, self.block.intern(name))
             self.block.bind_var(
                 self.writer, alias.asname or alias.name, member.expr)
+    elif node.module == '__future__':
+      for alias in node.names:
+        name = alias.name
+        if name in future_features:
+          (flag, implemented) = future_features[name]
```
trotterdylan: Style nit: kill unnecessary parens:

flag, implemented = future_features[name]

compiler/stmt.py
```diff
@@ -286,6 +309,21 @@ def visit_ImportFrom(self, node):
...
+          if not implemented:
+            msg = 'future feature {} not yet implemented by grumpy'.format(name)
+            raise util.ParseError(msg)
+          self.parser_flags |= flag
+        elif name in redundant_future_features:
+          # no-op
+          pass
+        else:
```
trotterdylan: Kill the no-op elif and replace it with:

elif name not in redundant_future_features:
   ... raise ParseError

compiler/stmt_test.py
```diff
@@ -273,12 +273,22 @@ def testImportNativeType(self):
...
         print Duration""")))
 
-  def testPrint(self):
+  def testPrintStatement(self):
     self.assertEqual((0, 'abc 123\nfoo bar\n'), _GrumpRun(textwrap.dedent("""\
         print 'abc',
         print '123'
         print 'foo', 'bar'""")))
 
+  def testFutureFeaturePrintFunction(self):
```
trotterdylan: Mind adding a test against one of the unimplemented future features?

runtime/builtin_types.go
```diff
@@ -431,6 +432,44 @@ func builtinOrd(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			}
+			sep = kwsep.Value()
+		case "end":
+			kwend, raised := ToStr(f, kwarg.Value)
+			if raised != nil {
+				return nil, raised
+			}
+			end = kwend.Value()
+		case "file":
+			// TODO: need to map Python sys.stdout, sys.stderr etc. to
```
trotterdylan: Please wrap comments at 80 chars.

kmod:
It's probably too late to change this, but I would definitely recommend operating on the Python bytecode rather than the Python ast.  There are a lot of subtleties in the ast->bytecode compilation process that take time to implement correctly, and I feel like we made a mistake in Pyston by operating on the ast.

compiler/stmt.py
```diff
@@ -41,6 +63,7 @@ def __init__(self, block_):
     self.block = block_
     self.writer = util.Writer()
     self.expr_visitor = expr_visitor.ExprVisitor(self.block, self.writer)
+    self.parser_flags = 0
```
paulsmith: I think we can get away with a single pass. According to [the 2.x docs](https://docs.python.org/2/reference/simple_stmts.html#future):

> A future statement must appear near the top of the module. The only lines that can appear before a future statement are:
> 
> * the module docstring (if any),
> * comments,
> * blank lines, and
> * other future statements.

None of which any future feature would alter the parsing logic of. So I think we can safely proceed with any flags after they are encountered.

compiler/stmt_test.py
paulsmith: Done.

runtime/builtin_types.go
```diff
@@ -431,6 +432,44 @@ func builtinOrd(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return NewInt(result).ToObject(), nil
 }
 
+func builtinPrint(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
```
paulsmith: Done.

compiler/stmt.py
trotterdylan: I'm guessing that the ast module doesn't check that `from __future__` imports satisfy these requirements so I think we should add a check and raise ParseError when `from __future__` appears later in the file.

Not sure the best way to do this. Maybe override NodeVisitor.visit()?

paulsmith: Yep -- that is what I came up with. I should have an update shortly.

paulsmith:
@trotterdylan I cleaned up some pylint errors from the code I added, but when I run `make precommit`, there are a number of errors for other parts of the code base I didn't edit. I'm not sure if there is something in my local environment that's causing pylint to not be configured correctly or something, but I'm not sure I can make the precommit pass cleanly as is.

compiler/stmt.py
```diff
@@ -41,6 +65,40 @@ def __init__(self, block_):
...
+    self.did_future_pass = False
+    self.future_lineno = 0
+
+  def visit(self, node):
+    root = node
+    # If this is the module node, do an initial pass through the module body's
```
trotterdylan: Sorry to bring this up late, but what do you think about having a separate FutureVisitor (or something) class that performs this logic ahead of time. The future flags can then be computed and passed into the ModuleBlock constructor (somewhere around [here](https://github.com/google/grumpy/blob/master/tools/grumpc#L53)) and accessed by the StatementVisitor.

The nice thing about that is it reduces the amount of state managed by the StatementVisitor. I'm concerned about increasing the complexity of StatementVisitor because it could easily become a catch-all and get out of control given its centrality.

trotterdylan:
@paulsmith I'm not sure what's up with the pylint errors. I don't see anything in a clean branch. Can you paste some of what you're seeing?

compiler/stmt.py
paulsmith: Sure, makes sense. I'll put something together.

paulsmith:
@trotterdylan Re lint errors: https://gist.github.com/paulsmith/683cc5b5fda04a556f07a5ef6a4042f2

trotterdylan:
@paulsmith I think pylint may be checking against Python 3 syntax. [This line](https://gist.github.com/paulsmith/683cc5b5fda04a556f07a5ef6a4042f2#file-make-precommit-log-L67) for example says a print statement is invalid syntax, which would be true for 3.

My guess is that pylint is executing under your system python install instead of python2.7. You could hardcode in the Makefile [here](https://github.com/google/grumpy/blob/master/Makefile#L161) what Python to execute pylint with.

This confusion has messed up the build for other folks as well. I'm going to add some checks to make sure the Python being used is always 2.7.

paulsmith:
@trotterdylan PTAL -- I refactored the future pass logic into a `visit_future` function, which returns an object that contains relevant future info and is passed to the StatementVisitor via ModuleBlock.

compiler/stmt.py
```diff
@@ -32,15 +32,106 @@
...
+
+# These future features are already in the language proper as of 2.6, so
+# importing them via __future__ has no effect.
+redundant_future_features = ["generators", "with_statement", "nested_scopes"]
+
+late_future = 'from __future__ imports must occur at the beginning of the file'
+
+
+def import_from_future(node):
+  """processes a future import statement, returning set of flags it defines."""
```
trotterdylan: Capitalize "processes..."

compiler/stmt.py
```diff
@@ -32,15 +32,106 @@
...
+
+
 class StatementVisitor(ast.NodeVisitor):
   """Outputs Go statements to a Writer for the given Python nodes."""
 
   # pylint: disable=invalid-name,missing-docstring
 
   def __init__(self, block_):
     self.block = block_
+    self.future_features = self.block.future_features or FutureFeatures()
```
trotterdylan: The pattern I've used for accessing global state like this (and I'm not saying it's a good pattern, just an established one) is to put a property on Block that pulls the corresponding attribute from self._module_block. See for example [full_package_name](https://github.com/google/grumpy/blob/master/compiler/block.py#L83).

future_features seems like a similar kind of global state. Can we use that pattern instead?

compiler/stmt.py
```diff
@@ -32,15 +32,106 @@
...
     self.writer = util.Writer()
     self.expr_visitor = expr_visitor.ExprVisitor(self.block, self.writer)
+    self.parser_flags = getattr(block_, 'parser_flags', 0)
```
trotterdylan: This no longer seems necessary since everyone accesses parser_flags from the future_features attr?

runtime/core.go
```diff
@@ -549,27 +550,34 @@ func Next(f *Frame, iter *Object) (*Object, *BaseException) {
 	return next.Fn(f, iter)
 }
 
-// Print implements the Python print statement. It calls str() on the given args
-// and outputs the results to stdout separated by spaces. Similar to the Python
-// print statement.
-func Print(f *Frame, args Args, nl bool) *BaseException {
-	// TODO: Support outputting to files other than stdout and softspace.
+// pyPrint encapsulates the logic of the Python print function.
+func pyPrint(f *Frame, args Args, sep, end string, file io.Writer) *BaseException {
```
trotterdylan: Please move this to the end of the file. I generally try to keep private helper functions at the end of the file. Also, maybe call it printImpl.

compiler/stmt.py
paulsmith: Done.

paulsmith:
@trotterdylan That did turn out to be the problem, re a Py3k version of pylint. Hard-coding to a 2.7 version made `precommit` clean.

trotterdylan:
This is great! Thanks for taking the time to push this through!

PR #25, opened 2017-01-05
Started implementing stdlib random

wuttem:
Implementation of some functions of the stdlib random functions.
These are the ones i needed for a test, but i could add some more.


wuttem:
I signed it!

lib/random.py
```diff
@@ -0,0 +1,28 @@
+"""Generate pseudo random numbers. Should not be used for security purposes"""
```
trotterdylan: Please add a period at the end of the end:

"""Generate pseudo random numbers. Should not be used for security purposes."""

lib/random.py
```diff
@@ -0,0 +1,28 @@
+"""Generate pseudo random numbers. Should not be used for security purposes."""
+
+from __go__.math.rand import Float64, Seed
+from __go__.time import Now
+
+
+# default seed with time like in cpython
+Seed(Now().UnixNano())
```
trotterdylan: Because Grumpy modules may be used by plain Go programs, I'm inclined to just rely on Go's default seeding behavior. This will result in a divergence in the way CPython and Grumpy programs execute, but issues that arise from the difference would be problems with the program, not with Grumpy.

lib/random.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def random():
+  return Float64()
+
+
+def randint(a, b):
+  if int(a) != a or int(b) != b:
+    raise ValueError("non-integer for randrange()")
+
```
trotterdylan: Please remove some of the empty lines. I prefer more compact code.

lib/random_test.py
```diff
@@ -0,0 +1,54 @@
...
+    item_idx = item/2
+    assert seq[item_idx] == item
+
+  try:
+    random.choice([])
+  except IndexError:
+    pass
+  else:
+    raise AssertionError("IndexError not raised")
+
```
trotterdylan: Two blank lines between functions, please

wuttem:
I looked at the cpython random lib. All functions are using a single random generator (_urandom from os module). I think we could use math.rand.Read from Go with a small wrapper as a replacment. Then we could basically copy the code from cpython once the needed math functions are implemented. Is this the way to go ? Is it possible to copy parts of the code or can we only copy whole files to /thirdparty?

ns-cweber:
@trotterdylan What's the strategy for using CPython standard library modules, provided they can be built by grumpy? Will grumpy ship with copies of these files, or will it try to load them from the system (or pyenv, etc) Python stdlib?

trotterdylan:
@wuttem Sorry, I lost track of this question. I'd rather copy whole files and remove (or comment out) parts we don't need. Long term I'd like to use the unmodified CPython .py sources.

@ns-cweber Grumpy will ship with them. You can copy them and drop them in third_party/stdlib. For consistency, please use files from [2.7.12rc1](https://www.python.org/downloads/release/python-2712rc1/).

trotterdylan:
@wuttem I like the idea of using math.rand.Read to implement a small wrapper.

wuttem:
I have a solution which would use math.rand.Read like described.
The most important Methods would work like in the cpython std lib if we have the following things implemented first:

- int.from_bytes
- int.bit_size
- RShift operation

I think it would be to ugly to shim these.

trotterdylan:
from_bytes and bit_length (assume you mean this rather than bit_size) seemed to have been introduced in Python 3. Is there a Python 2 equivalent that would work? struct maybe? Or could we use the Go encoding/binary package?

wuttem:
I saw that RShift is already working. I implemented small replacements for the needed integer functions. GrumpyRandom is a small Replacement for the urandom System Random Generator used in Cpython. The Random class is nearly 1:1 (code from cpython. The most needed functions are working. The remaining Code from cpythons random.py could be inserted once all requirements are implemented in the grumpy std lib.

lib/random.py
```diff
@@ -0,0 +1,187 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Generate pseudo random numbers. Should not be used for security purposes."""
+
+from __go__.math.rand import Uint32, Seed
+from __go__.math import Pow as _pow
```
trotterdylan: Any reason not to just import as Pow?

lib/random.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+BPF = 53  # Number of bits in a float
+RECIP_BPF = _pow(2, -BPF)
+
+
+def _gorandom(nbytes):
+  byte_arr = []
+  while len(byte_arr) < nbytes:
+    i = Uint32()
```
trotterdylan: I'm assuming we need to do this because we can't allocate a []byte to pass into Read()? Maybe put a TODO to mention that.

lib/random.py
```diff
@@ -0,0 +1,187 @@
...
+  i = 0
+  n = len(bytes) - 1
+  while n >= 0:
+    i += bytes[n] << (8 * n)
+    n -= 1
+  return i
+
+
+class GrumpyRandom(object):
+  """Alternate random number generator using math.rand.read as a replacement
```
trotterdylan: Please start docstrings with a one line summary per [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)

lib/random.py
```diff
@@ -0,0 +1,187 @@
...
+    return r
+
+  def getstate(self):
+    raise NotImplementedError('Entropy source does not have state.')
+
+  def setstate(self):
+    raise NotImplementedError('Entropy source does not have state.')
+
+
+# Most of the following code is taken from the cpython std lib.
```
trotterdylan: This is unfortunately a licensing problem because PSF licensed code needs to reside in third_party/stdlib. Is there a way we can reuse the CPython code as-is?

lib/random.py
wuttem: I took this convention from the cpython random.py code. all Math functions there are imported with leading underscores ... so it would fit to the others once the remaining code is inserted...

lib/random.py
wuttem: exactly, i will add the todo. one could implement a readBytes somewhere in go which returns a python list of integers.

lib/random.py
wuttem: There need to be some (really small) adjustments. Should i split the code in two files - one with the grumpy license and one under the cpython license ? i could make the minor adjustments under the cpython license ...

trotterdylan: Stuff that's from CPython should live in third_party/stdlib. Ideally we copy full sources from CPython and make light modifications in that dir. Code novel to Grumpy should live in lib. Does this give you enough to go on?

lib/_random.py
```diff
@@ -0,0 +1,113 @@
...
+    i = Uint32()
+    byte_arr.append(i & 0xff)
+    byte_arr.append(i >> 8 & 0xff)
+    byte_arr.append(i >> 16 & 0xff)
+    byte_arr.append(i >> 24 & 0xff)
+  byte_arr = byte_arr[0:nbytes]
+  return byte_arr
+
+
+def _notimplemented(*args, **kwargs):
```
trotterdylan: Move this to random.py since it's only used there.

lib/_random.py
```diff
@@ -0,0 +1,113 @@
...
+from __go__.math.rand import Uint32, Seed
+from __go__.math import Pow
+from __go__.time import Now
+
+
+BPF = 53  # Number of bits in a float
+RECIP_BPF = Pow(2, -BPF)
+
+
+# TODO
```
trotterdylan:
```
# TODO: The random byte generator currently...
```

third_party/stdlib/random.py
```diff
@@ -0,0 +1,645 @@
...
+    #             return self.sample(tuple(population), k)
+    #     return result
+
+## -------------------- real-valued distributions  -------------------
+
+## -------------------- uniform distribution -------------------
+
+    # def uniform(self, a, b):
+    #     "Get a random number in the range [a, b) or [a, b] depending on rounding."
+    #     return a + (b-a) * self.random()
```
trotterdylan: This can work, right?

wuttem:
Basics are ready. I added uniform. Maybe some more methods could already be added (i lost track of what is already in math and itertools but there is quite some progress). I think i will leave it like this for now and maybe after a 1-2 weeks i or someone else could come back to it and add all other methods at once with another PR.

PR #28, opened 2017-01-05
Add support for the `abs` builtin

meadori:
Implement the `abs` as defined here:

* https://docs.python.org/2.7/library/functions.html#abs

### DIFF
--- a/runtime/int.go
+++ b/runtime/int.go
@@ -66,6 +66,18 @@ func (i *Int) IsTrue() bool {
 // IntType is the object representing the Python 'int' type.
 var IntType = newBasisType("int", reflect.TypeOf(Int{}), toIntUnsafe, ObjectType)
 
+func intAbs(f *Frame, o *Object) (*Object, *BaseException) {
+	z := toIntUnsafe(o)
+	if z.Value() > 0 {
+		return z.ToObject(), nil
+	}
+	if z.Value() == MinInt {
+		nz := big.NewInt(int64(z.Value()))
+		return NewLong(nz.Neg(nz)).ToObject(), nil
+	}
+	return NewInt(-z.Value()).ToObject(), nil
+}
+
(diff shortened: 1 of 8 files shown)


meadori:
`make precommit` passed.

trotterdylan:
I realized that the right way to implement this is with an __abs__ slot
that's implemented by int, float etc.

runtime/int.go
```diff
@@ -66,6 +66,14 @@ func (i *Int) IsTrue() bool {
 // IntType is the object representing the Python 'int' type.
 var IntType = newBasisType("int", reflect.TypeOf(Int{}), toIntUnsafe, ObjectType)
 
+func intAbs(f *Frame, o *Object) (*Object, *BaseException) {
+	z := toIntUnsafe(o).Value()
+	if z <= 0 {
+		z = -z
```
trotterdylan: This can overflow if z == MinInt so check for that case and promote to Long.

runtime/int.go
```diff
@@ -66,6 +66,14 @@ func (i *Int) IsTrue() bool {
 // IntType is the object representing the Python 'int' type.
 var IntType = newBasisType("int", reflect.TypeOf(Int{}), toIntUnsafe, ObjectType)
 
+func intAbs(f *Frame, o *Object) (*Object, *BaseException) {
+	z := toIntUnsafe(o).Value()
+	if z <= 0 {
```
trotterdylan: To avoid an unnecessary allocation in the positive case, how about:

```
z := toIntUnsafe(o)
if z.Value() > 0 {
  return z.ToObject(), nil
}
// .. check for overflow as mentioned below
return NewInt(-x).ToObject(), nil
```

meadori:
Thanks for the review.  I address the copy for > 0 and overflow issues.

PR #31, opened 2017-01-05
Implemented dirname, exists, isdir, isfile in the os.path module

leekchan:
Added os.path.dirname, os.path.exists, os.path.isdir, os.path.isfile.

### DIFF
--- a/lib/os/path.py
+++ b/lib/os/path.py
@@ -15,6 +15,7 @@
 """"Utilities for manipulating and inspecting OS paths."""
 
 from __go__.path.filepath import Clean as normpath, IsAbs as isabs, Join, Dir as dirname  # pylint: disable=g-multiple-import,unused-import
+from __go__.os import Stat
 
 
 # NOTE(compatibility): This method uses Go's filepath.Join() method which
@@ -35,3 +36,24 @@ def join(*paths):
   if result and not paths[-1]:
     result += '/'
   return result
+
+
+def exists(path):
+  _, err = Stat(path)
+  if err is None:
+    return True
+  return False
+
+
+def isdir(path):
+  info, err = Stat(path)
+  if info and err is None:
+    return info.Mode().IsDir()
+  return False
+
+
+def isfile(path):
+  info, err = Stat(path)
+  if info and err is None:
+    return info.Mode().IsRegular()
+  return False
--- a/lib/os/path.py
+++ b/lib/os/path.py
@@ -14,8 +14,27 @@
 
 """"Utilities for manipulating and inspecting OS paths."""
 
-from __go__.path.filepath import Clean, IsAbs as isabs, Join, Dir as dirname  # pylint: disable=g-multiple-import,unused-import
 from __go__.os import Stat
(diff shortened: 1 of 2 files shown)


leekchan:
I signed it!

lib/os/path.py
```diff
@@ -35,3 +36,24 @@ def join(*paths):
   if result and not paths[-1]:
     result += '/'
   return result
+
+
+def exists(path):
```
trotterdylan: Would it be simpler to use [IsExist()](https://golang.org/pkg/os/#IsExist)?

lib/os/path.py
```diff
@@ -14,7 +14,8 @@
 
 """"Utilities for manipulating and inspecting OS paths."""
 
-from __go__.path.filepath import Clean as normpath, IsAbs as isabs, Join  # pylint: disable=g-multiple-import,unused-import
+from __go__.path.filepath import Clean as normpath, IsAbs as isabs, Join, Dir as dirname  # pylint: disable=g-multiple-import,unused-import
```
trotterdylan: I think we need a wrapper for Clean() because it doesn't behave the same way as normpath() when the input is unicode:

CPython:
```
>>> path.normpath(u'a/../b')
u'b'
```

Grumpy:
```
print repr(Clean(u'a/../b'))
'b'
```

lib/os/path_test.py
```diff
@@ -49,5 +51,43 @@ def TestNormPath():
...
+def TestDirname():
+  assert path.dirname('/a/b/c') == '/a/b'
+  assert path.dirname('/a/b/c/') == '/a/b/c'
+
+
+def TestExists():
+  _, file_path = tempfile.mkstemp()
+  dir_path = tempfile.mkdtemp()
+  try:
+    assert path.exists(file_path)
```
trotterdylan: Also check not exists. Ditto for the tests below.

lib/os/path.py
```diff
@@ -35,3 +36,33 @@ def join(*paths):
   if result and not paths[-1]:
     result += '/'
   return result
+
+
+def normpath(path):
+  result = Clean(path)
+  if isinstance(path, unicode):
+    # grumpy compiler encoded the string into utf-8,
```
trotterdylan: Please capitalize sentences in comments and wrap at 80 chars.

lib/os/path.py
```diff
@@ -35,3 +36,33 @@ def join(*paths):
   if result and not paths[-1]:
     result += '/'
   return result
+
+
+def normpath(path):
```
trotterdylan: Sorry for noticing this late, could you put the methods in this module in alphabetical order please? Also the tests in the test file.

lib/os/path.py
```diff
@@ -35,3 +36,33 @@ def join(*paths):
...
+  if isinstance(path, unicode):
+    # grumpy compiler encoded the string into utf-8,
+    # so the result can be decoded using utf-8.
+    return unicode(result, 'utf-8')
+  return result
+
+
+def exists(path):
+  _, err = Stat(path)
+  if err is None:
```
trotterdylan: return err is None

Issue #19, opened 2017-01-04
sys.{stdin,stdout,stderr} missing


JustinAzoff:
```
$ make run
import sys
sys.stdout.write("hi")
^D
AttributeError: 'module' object has no attribute 'stdout'
exit status 1
make: *** [run] Error 1
```

alanjds:
On master it seems fixed. Can you close the issue please?

trotterdylan:
Oh thanks for flagging @alanjds

PR #32, opened 2017-01-05
fix #30 and add #19 sys.stdin, sys.stdout, sys.stderr

### LINKED ISSUE #30: NewFileFromFD did not call fileInit which did not initialize file.reader causing fp.read(num) crash
when we open a file to read by fd, the NewFileFromFD in runtime/file.go will return a new File object file, but it did not call fileInit, due to this, the file.reader will be still nil, but then if we call file.read(num) to read num bytes, the in fileRead it will call file.reader.Read, and this will cause nil pointer dereference.

poc:

add a test method in sys module:

```
def test():
    # take stdin as an example
    fpd = NewFileFromFD(0)
    data = fpd.read(10)
    print data
```

then use grumpy run this:

```
import sys
sys.test()
```

got this error:

feilengcui008:


runtime/file.go
```diff
@@ -44,7 +44,15 @@ type File struct {
...
 	// TODO: Use fcntl or something to get the mode of the descriptor.
-	return &File{Object: Object{typ: FileType}, mode: "?", open: true, file: os.NewFile(fd, "<fdopen>")}
+	file := &File{Object: Object{typ: FileType}, mode: "?", open: true, file: os.NewFile(fd, "<fdopen>")}
+	initFileReader(file)
+	return file
+}
+
+func initFileReader(file *File) {
+	if file != nil && file.file != nil {
+		file.reader = bufio.NewReader(file.file)
```
trotterdylan: Generally I prefer to avoid helper functions unless they're encapsulating a lot of logic or they avoid some duplication. How about putting the file.reader initialization directly into the &File{...} assignment in NewFileFromFD()?

denismakogon: +1 on this

feilengcui008: ok, I'll add file.reader initilization into NewFileFromFD directly

trotterdylan:
Thanks for getting this done!

Issue #12, opened 2017-01-04
Can't access Go struct fields from Python


ns-cweber:
The following code fails with the error message: `AttributeError: '*Response' object has no attribute 'Body'`

```py
from __go__.os import Stdout as stdout
from __go__.net.http import DefaultClient as client

rsp, err = client.Get("http://www.google.com")
if err is not None:
    raise Exception(err.Error())

_, err = Copy(stdout, rsp.Body)
if err is not None:
    raise Exception(err.Error())
```

trotterdylan:
Thanks for submitting this. I think this used to work but the native interface has been rewritten a couple times and I think this feature got lost in the shuffle at some point. Will take a look.

PR #41, opened 2017-01-05
Issue #12: Fix native attribute access

ns-cweber:
This seems to fix issue #12, but I wouldn't be surprised if this fix is wrong. I'll add test cases, documentation, etc if this is indeed the correct fix.

### DIFF
--- a/runtime/native.go
+++ b/runtime/native.go
@@ -328,18 +328,6 @@ func getNativeType(rtype reflect.Type) *Type {
 			base = StrType
 		}
 		d := map[string]*Object{"__module__": builtinStr.ToObject()}
-
-		derefed := rtype
-		for derefed.Kind() == reflect.Ptr {
-			derefed = derefed.Elem()
-		}
-		if derefed.Kind() == reflect.Struct {
-			for i := 0; i < derefed.NumField(); i++ {
-				name := derefed.Field(i).Name
-				d[name] = newNativeField(name, i)
-			}
-		}
-
 		numMethod := rtype.NumMethod()
 		for i := 0; i < numMethod; i++ {
 			meth := rtype.Method(i)
@@ -349,7 +337,27 @@ func getNativeType(rtype reflect.Type) *Type {
 				d[meth.Name] = newNativeMethod(meth.Name, meth.Func)
 			}
 		}
-		t = &newNativeType(rtype, base, newStringDict(d)).Type
+
+		// Building a field for this type depends on the type itself, but we
+		// can't construct the type with the real `*Dict` yet, because we
+		// couldn't otherwise add the new fields to the `*Dict` (because we
+		// don't have access to a `*Frame` in this function). Instead, we'll
+		// convert `d` into a `*Dict` after we construct all of its fields, and
+		// we'll stick it back on `t` at that point.
+		t = &newNativeType(rtype, base, nil).Type
+
+		derefed := rtype
+		for derefed.Kind() == reflect.Ptr {
+			derefed = derefed.Elem()
+		}
+		if derefed.Kind() == reflect.Struct {
+			for i := 0; i < derefed.NumField(); i++ {
+				name := derefed.Field(i).Name
+				d[name] = newNativeField(name, i, t)
+			}
+		}
+		t.dict = newStringDict(d)
+
 		// This cannot fail since we're defining simple classes.
 		if err := prepareType(t); err != "" {
 			logFatal(err)
@@ -360,9 +368,9 @@ func getNativeType(rtype reflect.Type) *Type {
 	return t
 }
 
-func newNativeField(name string, i int) *Object {
+func newNativeField(name string, i int, t *Type) *Object {
 	nativeFieldGet := func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
-		if raised := checkFunctionArgs(f, name, args, nativeType); raised != nil {
+		if raised := checkFunctionArgs(f, name, args, t); raised != nil {
 			return nil, raised
 		}
 		v := toNativeUnsafe(args[0]).value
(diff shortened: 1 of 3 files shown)


ns-cweber:
I signed it!

ns-cweber:
I was concerned about copying over the field values onto the dict from a pointer type lest those values changed between copy time and access time. Maybe we could create a nativeFieldDescriptor or something and stick one of those on the dict for each public field? The descriptor would dereference the pointer at access time.

Note to self: Regardless of approach, I just realized I need to make sure this is only called on `*struct` types.

trotterdylan:
That sounds like a good approach to me. I think it's OK to support __get__ on struct, *struct, **struct, etc. But __set__ would only work if [CanSet()](https://golang.org/pkg/reflect/#Value.CanSet) is true for the field (support for __set__ could come later though.)

ns-cweber:
I added the aforementioned descriptors for accessors. It's ready for review when you get a chance. I'll work on the setters in the meanwhile.

runtime/native.go
```diff
@@ -348,6 +369,19 @@ func getNativeType(rtype reflect.Type) *Type {
...
+	// Closes over `i` and `refDepth`
+	nativeFieldGet := func(f *Frame, desc, instance *Object, owner *Type) (*Object, *BaseException) {
+		v := toNativeUnsafe(instance).value
+		for i := 0; i < refDepth; i++ {
+			v = v.Elem()
+		}
+		return WrapNative(f, v.Field(i))
+	}
+
+	return &Object{typ: &Type{slots: typeSlots{Get: &getSlot{nativeFieldGet}}}}
```
trotterdylan: I think you can use the existing `property` type for this:

```
return newProperty(
  newBuiltinFunction(fieldName, func(f *Frame, args Args, _ KWargs) (*Object, *BaseException) {
    if raised := checkFunctionArgs(f, fieldName, args, theNativeType); raised != nil {
      return nil, raised
    }
    v := toNativeUnsafe(args[0]).value
  }).ToObject(), nil, nil)
)
```

Note that I'm validating that arg[0] has the expected Python type before calling toNativeUnsafe(). You must only use the to*Unsafe() functions when you're sure that the argument is an instance of that type. Otherwise you're doing an unsafe cast between incompatible types (e.g. Str and Float).

runtime/native.go
```diff
@@ -348,6 +369,19 @@ func getNativeType(rtype reflect.Type) *Type {
 	return t
 }
 
+func newNativeField(name string, i, refDepth int) *Object {
```
trotterdylan: I wouldn't bother with the refDepth, just deref until you hit a non-Ptr type.

runtime/native.go
ns-cweber: It looks like args is always going to be the native instance because [propertyGet][0] always passes the instance argument, and the only instance we're putting this Get property on are native instances, right? I'm not seeing the scenario where this code could execute with a non-native instance...

[0]: https://github.com/google/grumpy/blob/master/runtime/descriptor.go#L65

trotterdylan: The property has an fget attribute which somebody in Python land could call with any old parameters (I think fget is not yet exposed by Grumpy, but it will be eventually).

ns-cweber: I see. At any rate, I included the check in my most recent push, which should be ready for review.

runtime/native.go
```diff
@@ -348,6 +360,21 @@ func getNativeType(rtype reflect.Type) *Type {
 	return t
 }
 
+func newNativeField(name string, i int) *Object {
+	nativeFieldGet := func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		if raised := checkFunctionArgs(f, name, args, nativeType); raised != nil {
```
trotterdylan: This check needs to verify that arg[0] is the particular type being created, not just any native type. If this getter is created for *SomeStruct and I call fget with, say, *int then it will pass this check (it is a native type) but the v.Field() call below will panic.

This may require a small reorganization where you create the type before fully populating its dict in getNativeType() above. Something like:

```
t = &newNativeType(rtype, base, newStringDict(d)).Type
for each field {
  raised := t.dict.SetItemString(f, fieldName, newNativeField(fieldName, fieldIndex, t))
  ...
}
```

ns-cweber:
Alright, I added the type check and a corresponding unit test.

runtime/native_test.go
```diff
@@ -413,6 +413,33 @@ func TestNativeTypeName(t *testing.T) {
 	}
 }
 
+func TestNewNativeField_ChecksInstanceType(t *testing.T) {
```
trotterdylan: Please remove the underscore in the test name. It's a bit bulky but I don't think underscores have yet been used elsewhere in test names.

runtime/native.go
```diff
@@ -337,7 +337,27 @@ func getNativeType(rtype reflect.Type) *Type {
...
 		}
-		t = &newNativeType(rtype, base, newStringDict(d)).Type
+
+		// Building a field for this type depends on the type itself, but we
+		// can't construct the type with the real `*Dict` yet, because we
+		// couldn't otherwise add the new fields to the `*Dict` (because we
+		// don't have access to a `*Frame` in this function). Instead, we'll
+		// convert `d` into a `*Dict` after we construct all of its fields, and
+		// we'll stick it back on `t` at that point.
+		t = &newNativeType(rtype, base, nil).Type
```
trotterdylan: You can remove the dict param from newNativeType(). The function is used in one other spot in native_test.go and I think you'll have to manually populate the dict field on the result.

ns-cweber: The tests pass without manually setting the `dict`, so I left it off.

bitmoji

trotterdylan:
This is great! Thanks for addressing all the feedback and getting this done!

PR #44, opened 2017-01-05
Implement namedtuple with type()

selik:
Grumpy does not allow exec'ing a class definition. This implementation
uses dynamic class creation via the ``type`` function. The
``NamedTuple`` class definition mimics the original interpolated class
definition template. The ``namedtuple`` function now needs to insert
attributes and properties into the class dictionary before the class is
created rather than interpolating those same things.

After the class is created, some method docstrings must be updated.

This code was tested using the rough test code in the main section, not
with a full battery of unit tests.

### DIFF
--- a/third_party/stdlib/collections.py
+++ b/third_party/stdlib/collections.py
@@ -258,29 +258,35 @@ class OrderedDict(dict):
 ### namedtuple
 ################################################################################
 
-_class_template = '''\
-class {typename}(tuple):
+class NamedTuple(tuple):
     '{typename}({arg_list})'
 
     __slots__ = ()
 
-    _fields = {field_names!r}
+    _fields = ()
 
-    def __new__(_cls, {arg_list}):
+    def __new__(_cls, *args, **kwds):
         'Create new instance of {typename}({arg_list})'
-        return _tuple.__new__(_cls, ({arg_list}))
+        try:
+            kwargs = tuple(map(kwds.pop, list(_cls._fields[len(args):])))
+        except KeyError as e:
+            raise TypeError('Unexpected keyword argument %r' % e.message)
+        return tuple.__new__(_cls, args + kwargs)
 
     @classmethod
     def _make(cls, iterable, new=tuple.__new__, len=len):
         'Make a new {typename} object from a sequence or iterable'
         result = new(cls, iterable)
-        if len(result) != {num_fields:d}:
-            raise TypeError('Expected {num_fields:d} arguments, got %d' % len(result))
+        n = len(cls._fields)
+        if len(result) != n:
+            raise TypeError('Expected %d arguments, got %d' % (n, len(result)))
         return result
 
     def __repr__(self):
         'Return a nicely formatted representation string'
-        return '{typename}({repr_fmt})' % self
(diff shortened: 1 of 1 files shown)


selik:
@googlebot 
I signed it!

third_party/stdlib/collections.py
```diff
@@ -717,37 +713,36 @@ def __and__(self, other):
...
-#    print Point(11, 22)._replace(x=100)
-#
-#    Point3D = namedtuple('Point3D', Point._fields + ('z',))
-#    print Point3D.__doc__
-#
-#    import doctest
-#    TestResults = namedtuple('TestResults', 'failed attempted')
-#    print TestResults(*doctest.testmod())
+   # verify that instances can be pickled
+   from cPickle import loads, dumps
```
trotterdylan: cPickle will not work. I think you can just comment out lines 717 to 720 for now.

third_party/stdlib/collections.py
```diff
@@ -258,37 +258,43 @@ def viewitems(self):
...
 
-    _fields = {field_names!r}
+    _fields = ()
 
-    def __new__(_cls, {arg_list}):
+    def __new__(_cls, *args, **kwds):
         'Create new instance of {typename}({arg_list})'
-        return _tuple.__new__(_cls, ({arg_list}))
+        try:
+            kwargs = tuple(map(kwds.pop, list(_cls._fields[len(args):])))
```
trotterdylan: This line is breaking my brain. Can you break this into multiple statements?

third_party/stdlib/collections.py
```diff
@@ -258,37 +258,43 @@ def viewitems(self):
...
+        n = len(cls._fields)
+        if len(result) != n:
+            raise TypeError('Expected %d arguments, got %d' % (n, len(result)))
         return result
 
     def __repr__(self):
         'Return a nicely formatted representation string'
-        return '{typename}({repr_fmt})' % self
+        values = (getattr(self, name) for name in self._fields)
+        attrstring = ', '.join('%s=%r' % (k, v) for k, v in zip(self._fields, values))
```
trotterdylan: Slight less verbose: `', '.join('%s=%r' % kv for kv in zip(self._fields, values))`

third_party/stdlib/collections.py
selik: Sure. Instead of:

        try:
        	kwargs = tuple(map(kwds.pop, list(_cls._fields[len(args):])))
        except KeyError as e:
            raise TypeError('Unexpected keyword argument %r' % e.message)
        return tuple.__new__(_cls, args + kwd_values)

More clear?

        try:
        	keywords = list(_cls.fields[len(args):])
        	kwd_values = tuple(map(kwds.pop, keywords))
        except KeyError as e:
            raise TypeError('Unexpected keyword argument %r' % e.message)
        return tuple.__new__(_cls, args + kwd_values)

selik:
Finishing the builtins like ``map`` and ``all`` is certainly more important. I'll leave this alone for now.

trotterdylan:
I'm going to close this one for now. Feel free to reopen once we're more ready for it.

PR #45, opened 2017-01-05
dict: 64bit atomic operations must be 8-byte aligned.

nairb774:
Atomic operations on non-8 byte aligned values results in segfaults on some
32bit architectures. This aligns all 64 atomic operations as well as places
compile time checks to prevent regression.

### DIFF
--- a/runtime/dict.go
+++ b/runtime/dict.go
@@ -202,14 +202,14 @@ func (t *dictTable) writeEntry(f *Frame, index int, entry *dictEntry) (*dictTabl
 // dictEntryIterator is used to iterate over the entries in a dictTable in an
 // arbitrary order.
 type dictEntryIterator struct {
-	table *dictTable
 	index int64
+	table *dictTable
 }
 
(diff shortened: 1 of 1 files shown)


trotterdylan:
How does that work? Is the const expected to be unsigned and if it's negative cause a compilation error? Is there precedent for this in other code that we could reference?

nairb774:
Because unsafe.Offset returns a uintptr, the `-` cause a compiler error (negative can't be assigned to unsigned or some such) when the offset is off. I'm not sure if there is a precedent other than fixing it and hopefully documenting it somewhere. I tried to make the failure a little obvious by adding a comment to the line. Much better IMHO than the random panics I was seeing which didn't say anything about alignment.

I could also see this being in a unit test/test code, but it would have the chance of being separated from the variables and this seemed like an easy way to place the "requirement" near the code which is placing the requirement.

trotterdylan:
SGTM. Can you just add a more detailed comment describing the intent and mechanism?

PR #47, opened 2017-01-05
Don't trigger segfaults/panics when producing a string from an Object.

nairb774:
While trying to debug an issue in the native support, I was running with GDB, and kept getting hit with segfaults when (*Object)(nil)s were getting String() called on them (buried deep in a call to fmt.SomethingOrOther). This seems like a sensible return value in this case. I could also see arguing that this shouldn't happen, and the source should be addressed. I didn't get a stack trace when I ran into this, bit it shouldn't be too hard if we would rather go that route.

### DIFF
--- a/runtime/object.go
+++ b/runtime/object.go
@@ -71,6 +71,9 @@ func (o *Object) Dict() *Dict {
 
 // String returns a string representation of o, e.g. for debugging.
 func (o *Object) String() string {
+	if o == nil {
+		return "nil"
+	}
 	s, raised := Repr(newFrame(nil), o)
 	if raised != nil {
 		return fmt.Sprintf("<%s object (repr raised %s)>", o.typ.Name(), raised.typ.Name())


nairb774:
Sample (possibly test related - still on the fence for this fix):

```
#0  _/home/pi/git/grumpy/runtime.Repr (f=0x10c9d140, o=0x0, ~r2=0x0, ~r3=0x10) at /home/pi/git/grumpy/runtime/core.go:578
#1  0x000c9710 in _/home/pi/git/grumpy/runtime.(*Object).String (o=0x0, ~r0=<error reading variable: Cannot access memory at address 0x1>) at /home/pi/git/grumpy/runtime/object.go:74
#2  0x00287604 in fmt.(*pp).handleMethods (p=0x10c86000, verb=118, handled=true) at /home/pi/opt/go-1.7.3/src/fmt/print.go:590
#3  0x002878a0 in fmt.(*pp).printArg (p=0x10c86000, arg=..., verb=118) at /home/pi/opt/go-1.7.3/src/fmt/print.go:665
#4  0x0028cac8 in fmt.(*pp).doPrintf (p=0x10c86000, format="frame.Raise(%v, %v, %v)", a= []interface {} = {...}) at /home/pi/opt/go-1.7.3/src/fmt/print.go:985
#5  0x0028365c in fmt.Sprintf (format="frame.Raise(%v, %v, %v)", a= []interface {} = {...}, ~r2="") at /home/pi/opt/go-1.7.3/src/fmt/print.go:196
#6  0x00151624 in _/home/pi/git/grumpy/runtime.TestFrameRaise (t=0x10c86500) at /home/pi/git/grumpy/runtime/frame_test.go:139
#7  0x00092d00 in testing.tRunner (t=0x10c86500, fn={void (struct testing.T *)} 0x10c50fd4) at /home/pi/opt/go-1.7.3/src/testing/testing.go:610
#8  0x00072d9c in runtime.goexit () at /home/pi/opt/go-1.7.3/src/runtime/asm_arm.s:998
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

Issue #42, opened 2017-01-05
TypeError: cannot convert int to io.Writer (ioutil.Discard)


ns-cweber:
Trying to pass `ioutil.Discard` as an `io.Writer` argument raises a TypeError. Here's a simple example program:

```py
from __go__.io.ioutil import Discard as devnull
from __go__.fmt import Fprintln as fprintln

fprintln(devnull, "Hello, world")
```

The definition for `ioutil.Discard` is `var Discard io.Writer = devNull(0)`

trotterdylan:
I took a stab at this. The basic problem is that for primitive types, `__native__` is returning a reflect.Value holding the basic type of the underlying storage, not the "real" type. Since devNull is defined as `type devNull int`, the reflect.Value for the Python devNull type `<type devNull>` is a plain old int, not a Go devNull.

The most obvious fix is to have `<type devNull>` define it's own __native__ that fetches the basic type and then converts it to devNull. I'm not happy with the way the code's structured so I'm going to think about how to rework it, but https://github.com/google/grumpy/pull/52 outlines the basic approach.

trotterdylan:
This now works as expected.

PR #52, opened 2017-01-06
First pass at returning proper types from __native__ for primitive types

trotterdylan:
Attempt to address: https://github.com/google/grumpy/issues/42

### DIFF
--- a/runtime/native.go
+++ b/runtime/native.go
@@ -144,8 +144,12 @@ func nativeBoolNative(f *Frame, o *Object) (reflect.Value, *BaseException) {
 
 func nativeBoolNew(f *Frame, t *Type, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	meta := toNativeBoolMetaclassUnsafe(t.ToObject())
-	if len(args) == 0 {
+	argc := len(args)
+	if argc == 0 {
 		return meta.falseValue, nil
+	}
+	if argc != 1 {
+		return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("%s() takes at most 1 argument (%d given)", t.Name(), argc))
 	}
 	ret, raised := IsTrue(f, args[0])
 	if raised != nil {
--- a/runtime/native.go
+++ b/runtime/native.go
@@ -25,10 +25,11 @@ import (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


runtime/native.go
```diff
@@ -338,6 +338,13 @@ func getNativeType(rtype reflect.Type) *Type {
 			}
 		}
 		t = &newNativeType(rtype, base, newStringDict(d)).Type
+		t.slots.Native = &nativeSlot{func(f *Frame, o *Object) (reflect.Value, *BaseException) {
+			val, raised := base.slots.Native.Fn(f, o)
+			if raised != nil {
+				return reflect.Value{}, raised
+			}
+			return val.Convert(toNativeMetaclassUnsafe(o.typ.ToObject()).rtype), nil
+		}}
```
ns-cweber: Won't this overwrite slots for the primitive types which have their own Natives?

trotterdylan: It'll override the method of the primitive base type for the subclass t. The override calls the base implementation to get the primitive object and then coverts it to the (compatible) subtype.

E.g. `type devNull int` produces the Python `<type devNull>`. Without the override, the native object returned from a `<type devNull>` is the plain Go int type. The override converts this to the Go devNull type. From the original issue, devNull satisfies the `Writer` interface, whereas int does not.

ns-cweber: Cool. This was my only concern. Does this merit a test case?

trotterdylan:
Unfortunately this got a little more involved because of bool types. They're tricky because they're singletons. The trouble happens when you try to create a new bool instance of the correct Go subtype, you get the result of bool.__new__ which always returns the True or False singletons, not a new bool instance of the requested type.

I think the best way to deal with it is to handle bools specially by having a new metaclass that holds the singletons for each Go bool typedef. For example, `type myBool bool` will have an associated `<type myBool>` whose __new__ returns true and false singletons specific to that type.

This means that the following are true:

```
bool(1) == myBool(1)
bool(1) is bool("nonzero")
myBool(1) is myBool("nonzero")
bool(1) is not myBool(1)
```

TL;DR: each bool typedef defined in Go has its own true and false values that don't share identity with the usual Python True and False.

ns-cweber:
I think this makes sense.

PR #54, opened 2017-01-06
Add support for the `all` builtin

meadori:
Implement the `all` as defined here:

* https://docs.python.org/2/library/functions.html#all

### DIFF
--- a/runtime/builtin_types.go
+++ b/runtime/builtin_types.go
@@ -190,6 +190,31 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return Abs(f, args[0])
 }
 
+func builtinAll(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkFunctionArgs(f, "all", args, ObjectType); raised != nil {
+		return nil, raised
+	}
+	iter, raised := Iter(f, args[0])
+	if raised != nil {
+		return nil, raised
+	}
+	o, raised := Next(f, iter)
+	for ; raised == nil; o, raised = Next(f, iter) {
+		ret, raised := IsTrue(f, o)
+		if raised != nil {
+			return nil, raised
+		}
+		if !ret {
+			return False.ToObject(), nil
+		}
+	}
+	if !raised.isInstance(StopIterationType) {
+		return nil, raised
+	}
+	f.RestoreExc(nil, nil)
+	return True.ToObject(), nil
+}
+
 func builtinBin(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	if raised := checkFunctionArgs(f, "bin", args, ObjectType); raised != nil {
 		return nil, raised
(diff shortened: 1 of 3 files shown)


runtime/builtin_types.go
```diff
@@ -190,6 +190,29 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	}
+	iter, raised := Iter(f, args[0])
+	if raised != nil {
+		return nil, raised
+	}
+	o, raised := Next(f, iter)
+	for ; raised == nil; o, raised = Next(f, iter) {
+		if ret, raised := IsTrue(f, o); raised != nil {
+			return nil, raised
+		} else if !ret {
```
trotterdylan: Prefer to avoid else when possible.

```
if ret, raised := ... {
  return ...
}
if !ret {
  return ...
}
```

runtime/builtin_types.go
```diff
@@ -190,6 +190,29 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			return False.ToObject(), nil
+		}
+	}
+	if !raised.isInstance(StopIterationType) {
+		return nil, raised
+	}
+	f.RestoreExc(nil, nil)
```
trotterdylan: It's unfortunate that there's so much boilerplate for iterating over a sequence. [seqForEach](https://github.com/google/grumpy/blob/master/runtime/seq.go#L128) almost does what you want but doesn't support breaking out of the loop early. We could add support for that by returning an additional bool value from the callback which signals whether to continue. Might be messy though. I'll leave it up to you whether you want to try something like that.

runtime/builtin_types.go
ns-cweber: He couldn't use that syntax. He would have to declare `ret` outside of the `if`:

``` go
ret, raised := IsTrue(f, o)
if raised != nil {
    return nil, raised
}
if !ret {
    return False.ToObject(), nil
}
```

trotterdylan: Thanks for pointing that out. The code I wrote isn't correct for that reason. But declaring outside the if is the style I've followed in the source code. It is more verbose (and also pollutes the outer scope's namespace more) but I think the clarity is worth it.

meadori: > But declaring outside the if is the style I've followed in the source code.

I will change it to early return, but for the record I lifted that pretty much straight from `Contains` in `core.go` :smile:

runtime/builtin_types.go
meadori: I will try and see if it gets messy.  Thanks for the review!

runtime/builtin_types.go
trotterdylan: ... declaring outside the if is the style I've *tried* to follow in the source code... :D

runtime/builtin_types.go
meadori: Another issue is that `seqForEach` itself **only** returns the exception.  It would have to be modified to return an `*Object` as well.  Worse than that really -- an object for the false case and one from the true case (and those would probably need to be parameters as well).

trotterdylan: Ugh. Yeah. I think I remember trying to do this at some point and ran into the same thing. Let's leave things as-is and address this down the road with a more general mechanism.

PR #63, opened 2017-01-07
Add support for the `any` builtin (#54)

meadori:
Implement the `any` as defined here:

    * https://docs.python.org/2/library/functions.html#any


choleraehyq:
Haha, I'm about to send a similar PR.
LGTM.

runtime/builtin_types.go
```diff
@@ -637,3 +627,31 @@ func numberToBase(prefix string, base int, o *Object) string {
...
 	return prefix + s
 }
+
+// quantifyIterable implements the builtins "all" and "any".
+// It takes a condition function, "cond", that determines when
+// the iteration should stop.  If the iteration stops before
+// all the items are seen, then "condTrue" is returned. Otherwise,
+// "condFalse" is returned. For more details see:
+// https://docs.python.org/2/library/functions.html#all
+func quantifyIterable(f *Frame, o *Object, cond func(bool) bool, condTrue *Object, condFalse *Object) (*Object, *BaseException) {
```
trotterdylan: How about something like this:

```
func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) bool) (bool, *BaseException)
```

Then you can define predicates `seqFindFirstTrue()` and `seqFindFirstFalse()` to be used by `any` and `all`. `seqFindFirst()` can then be reused by `[grumpy.Contains()](https://github.com/google/grumpy/blob/master/runtime/core.go#L71)`.

What do you think?

trotterdylan: Actually the predicate would have to be able to return an exception, so:

```
func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) (bool, *BaseException)) (bool, *BaseException)
```

meadori: I like it!  Thanks for the suggestion.  I will have a new patch shortly.

trotterdylan:
Thanks for doing this! LGTM

PR #66, opened 2017-01-07
make itertools.chain be a class and from_iterable a class_method of c…

IuryAlves:
In python `itertools.chain` is implemented as a class and `from_iterable` is a classmethod.

### DIFF
--- a/lib/itertools.py
+++ b/lib/itertools.py
@@ -17,10 +17,19 @@
 import _collections
 import sys
 
-def chain(*iterables):
-  for it in iterables:
-    for element in it:
-      yield element
+class chain(object):
+
+    def __init__(self, *iterables):
+        self.iterables = iterables
+
+    @classmethod
+    def from_iterable(cls, iterable):
+        return cls(*iterable)
+
+    def __iter__(self):
+        for it in self.iterables:
+            for element in it:
+                yield element
 
 def compress(data, selectors):
   return (d for d,s in izip(data, selectors) if s)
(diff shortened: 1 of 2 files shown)


trotterdylan:
Thanks for the PR! The test is failing because neither classmethod nor decorators are implemented yet. Decorators is a bit involved but I think classmethods aren't too bad, and once we have them this change can be made to work this way:

```
def from_iterable(...):
  ...
from_iterable = classmethod(from_iterable)
```

I've filed issues https://github.com/google/grumpy/issues/72 and https://github.com/google/grumpy/issues/73

IuryAlves:
Thanks @trotterdylan

trotterdylan:
classmethod is now supported so this PR could work if you replace `@classmethod` with `from_iterable = classmethod(from_iterable)`

lib/itertools.py
```diff
@@ -17,15 +17,16 @@
 import _collections
 import sys
 
+def _from_iterable(cls, iterable):
```
S-YOU: This two lines supposed to be inside chain class, I think. 
Something like 
https://docs.python.org/2/whatsnew/2.4.html#pep-318-decorators-for-functions-and-methods

Any issue with inside the class?

IuryAlves: Nope. I just forgot that i could implement `from_iterable` inside the class.

kit1978:
> @@ -17,15 +17,16 @@
 import _collections
 import sys
 
+def _from_iterable(cls, iterable):This two lines supposed to be inside chain class, I think.
Something like

kit1978:
ขอ⁣ขอบคุณพี่ๆทุกท่านที่ช่วยเหลือครับ

ส่งจาก BlueMail ​

ใน 11 ม.ค. 2017, 20:22 ที่ 20:22 YOU <notifications@github.com> เขียน:

trotterdylan:
Looks great. Thanks for getting this done!

PR #67, opened 2017-01-08
Update README.md

muxuezi:
The url `http://jyni.org/` need `http`


trotterdylan:
Thanks for submitting this PR. Seems like this is now a no-op since another PR fixed the scheme of the URL, so I'm closing this one.

PR #68, opened 2017-01-08
Added os.chdir, os.getcwd, os.path.abspath.

leekchan:
Added os.chdir, os.getcwd, os.path.abspath.


trotterdylan:
Thanks for the PR! This looks great. Merging.

PR #69, opened 2017-01-08
Fix link to jyni.org in README.md

FlorianLudwig:
The link was interpreted as local and did link to https://github.com/google/grumpy/blob/master/jyni.org instead of http://jyni.org


FlorianLudwig:
Well, that was somewhat overkill to sign an CLA to fix a link but here you go,  I signed ;)

trotterdylan:
Thanks for the fix!

PR #80, opened 2017-01-09
Add support for the `cmp` builtin (#54) (#63)

meadori:
Implement the `cmp` as defined here:

* https://docs.python.org/2/library/functions.html#cmp

Implementing this builtin required implementing a fair
amount of supporting logic to do 3-way comparisons. The
3-way comparison implementation in CPython is somewhat
complex, thus there may be some cases that still need
support. This is a good start, though.

### DIFF
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -65,6 +65,29 @@ func Assert(f *Frame, cond *Object, msg *Object) *BaseException {
 	return raised
 }
 
+// compare implements a 3-way comparison which returns:
+//
+//   -1 if v < w
+//    0 if v == w
+//    1 if v > w
+//
+// It closely resembles the behavior of CPython's do_cmp in object.c.
+func Compare(f *Frame, v, w *Object) (*Object, *BaseException) {
+	cmp := v.typ.slots.Cmp
+	if v.typ == w.typ && cmp != nil {
+		return cmp.Fn(f, v, w)
+	}
+	r, raised := tryRichTo3wayCompare(f, v, w)
+	if r != NotImplemented {
+		return r, raised
+	}
+	r, raised = try3wayCompare(f, v, w)
+	if r != NotImplemented {
+		return r, raised
+	}
+	return NewInt(compareDefault(f, v, w)).ToObject(), nil
+}
+
 // Contains checks whether value is present in seq. It first checks the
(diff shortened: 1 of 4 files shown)


runtime/core.go
```diff
@@ -936,6 +940,106 @@ func compareDefault(f *Frame, v, w *Object) int {
...
+
+// compare implements a 3-way comparison which returns:
+//
+//   -1 if v < w
+//    0 if v == w
+//    1 if v > w
+//
+// It closely resembles the behavior of CPython's do_cmp in object.c.
+func compare(f *Frame, v, w *Object) (*Object, *BaseException) {
+	cmp := v.typ.slots.Cmp
```
trotterdylan: Can we just move this logic directly into `Compare()`?

testing/builtin_test.py
```diff
@@ -101,3 +101,133 @@ def __call__(self, *args, **kwargs):
 assert callable(bar)
 assert callable(bar())
 
+# cmp(x)
```
trotterdylan: Thanks for the thorough tests!

runtime/core.go
meadori: Will do.

meadori:
The rich compare implementation will need to be tweak to fall-back on 3-way compare if the appropriate rich comparison operation is **not** available.  I will do that in a separate PR.

meadori:
I made the one change you requested and will commit to adding the fallback to 3-way compare for rich comparisons in a follow-on PR.  Everything else look okay?

runtime/core.go
```diff
@@ -936,6 +959,83 @@ func compareDefault(f *Frame, v, w *Object) int {
...
+	if raised != nil {
+		return false, raised
+	}
+	return br, raised
+}
+
+// halfCompare tries a comparison with the __cmp__ slot, ensures the result
+// is an integer, and returns it. It closely resembles the behavior of CPython's
+// half_compare in typeobject.c.
+func halfCompare(f *Frame, v, w *Object) (*Object, *BaseException) {
```
trotterdylan: In a subsequent PR, would you mind changing the return value to (*Int, *BaseException)? You pass the result to intNeg() later on in the code and it would be easier to understand if it was an *Int.

trotterdylan:
Thanks again for all your work on this!

Issue #77, opened 2017-01-09
Missing support for string slicing


frankban:
This error is returned at runtime: `TypeError: string indices must be integers, not slice`.

ns-cweber:
I reproduced with `print("hello"[1:-1])`.

trotterdylan:
Thanks for reporting this. Proper slice support was never implemented for str and unicode, although it is implemented for tuples and lists [here](https://github.com/google/grumpy/blob/master/runtime/seq.go#L170). It should be a simple matter of handling slice type arguments for str [here](https://github.com/google/grumpy/blob/master/runtime/str.go#L204) and similarly for unicode.

frankban:
Thanks for the quick fix!

PR #83, opened 2017-01-10
Start implement string slicing

S-YOU:
This is for #77.  Once this is ok, I could continue unicode.

### DIFF
--- a/runtime/str.go
+++ b/runtime/str.go
@@ -201,9 +201,8 @@ func strGE(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return strCompare(v, w, False, True, True), nil
 }
 
-// strGetItem returns a slice of string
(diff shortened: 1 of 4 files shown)


runtime/str.go
```diff
@@ -201,20 +201,38 @@ func strGE(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return strCompare(v, w, False, True, True), nil
 }
 
+// strGetItem returns a slice of string
```
trotterdylan: Please wrap comments at 80 chars

S-YOU: @trotterdylan, fix that, and also added unicode slicing, should I use separate PR?

trotterdylan: @S-YOU No, it's fine to leave it all in one PR. Thanks!

Issue #73, opened 2017-01-09
Support classmethod


trotterdylan:
https://docs.python.org/2/library/functions.html#classmethod

S-YOU:
I am wondering about `from_iterable = classmethod(from_iterable)` approach,
Isn't `classmethod` need to know which class it is in?

Usage in `chain.from_iterable` or `UserDict.fromkeys`, 
both try to instantiate class instance right inside those functions, 
which means you need to pass Class itself as first parameter.

trotterdylan:
`classmethod` is a pretty straightforward application of the descriptor protocol. The class is passed as the last parameter of the `__get__` method:

```
>>> class Desc(object):
...   def __get__(self, obj, cls):
...     print cls
...
>>> class C(object):
...   foo = Desc()
...
>>> C.foo
<class '__main__.C'>
```

Here's CPython's implementation: https://github.com/python/cpython/blob/2.7/Objects/funcobject.c#L586

PR #86, opened 2017-01-10
Start implementing classmethod

S-YOU:
Working on #73, let me know if my approach is not right.
Tests still have to be done, can't figure out how to write it in go yet.

Manually Testing in python seems to be working
```
import UserDict
print UserDict.UserDict.fromkeys("abcde")
{'a': None, 'c': None, 'b': None, 'e': None, 'd': None}
```

### DIFF
--- a/runtime/function.go
+++ b/runtime/function.go
@@ -25,6 +25,9 @@ var (
 	// StaticMethodType is the object representing the Python
 	// 'staticmethod' type.
 	StaticMethodType = newBasisType("staticmethod", reflect.TypeOf(staticMethod{}), toStaticMethodUnsafe, ObjectType)
+	// ClassMethodType is the object representing the Python
+	// 'classmethod' type.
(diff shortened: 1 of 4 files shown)


runtime/function.go
```diff
@@ -182,3 +185,46 @@ func initStaticMethodType(map[string]*Object) {
...
+func (m *classMethod) ToObject() *Object {
+	return &m.Object
+}
+
+func classMethodGet(f *Frame, desc, _ *Object, typ *Type) (*Object, *BaseException) {
+	m := toClassMethodUnsafe(desc)
+	if m.callable == nil {
+		return nil, f.RaiseType(RuntimeErrorType, "uninitialized classmethod object")
+	}
+	return newBuiltinFunction("_classMethodGet", func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
```
trotterdylan: I think you want to return a [method object](https://github.com/google/grumpy/blob/master/runtime/method.go) rather than a function.

S-YOU: @trotterdylan, submitted new one. much better now.

Issue #11, opened 2017-01-04
library path includes "."


waitingkuo:
How to import libraries when its path contains "." ?

For example, I can import "github.com/labstack/echo" in pure go by

    import "github.com/labstack/echo"

In grumpy, I can import it like

    from __go__.github.com.labstack.echo import New

But it will be compiled to something like this

    import "github/com/labstack/echo"

Anyway to correct this issue?

trotterdylan:
I think long term I'll need an alternative import syntax for Go imports, like:

from __go__ import "github.com/labstack/echo"

But unfortunately that's not workable at the moment since I'm using the ast module for Python parsing and it won't accept this kind of syntax.

In the short term, perhaps something like:

New = __goimport__("github.com/labstack/echo", "New")

I'll put something together.

sebgoa:
Very dirty but, couldn't you temporarily hack grumpc to just detect `.com` and don't replace it with a `/`.

I am not familiar with ast, so I am going through the grumpc code now.

but yes , best would be `from __go__ import "github.com/labstack/echo"`

S-YOU:
golang.org does not include .com, I would suggest putting "github.com" and "golang.org" as special case for the time being.

PR #87, opened 2017-01-10
work-around for go module import

S-YOU:
This is a workaround for #11, go module import from github.com and golang.org
What do you think? @trotterdylan 

PS: I havn't checked tests yet.


S-YOU:
@trotterdylan, added comments, linked to vcs.go, changed name DOMAINS to VCS, added more VCS domains.

PR #88, opened 2017-01-11
Add support for non-standard Python install in make

trotterdylan:
Errors out when python is not 2.7.x with: unsupported Python version 3.6.0, Grumpy only supports 2.7.x. To use a different python binary such as python2, run: 'make PYTHON=python2 ...'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,14 @@ GOOS ?= $(word 1,$(GO_ENV))
 GOARCH ?= $(word 2,$(GO_ENV))
 ROOT_DIR := $(realpath .)
 PKG_DIR := build/pkg/$(GOOS)_$(GOARCH)
+PYTHON ?= python
+PYTHON_BIN := $(shell which $(PYTHON))
+PYTHON_VER := $(word 2,$(shell $(PYTHON) -V 2>&1))
 PY_DIR := build/lib/python2.7/site-packages
+
+ifeq ($(filter 2.7.%,$(PYTHON_VER)),)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -90,30 +97,31 @@ precommit: cover gofmt lint test
 $(COMPILER_BIN) $(RUNNER_BIN) $(TOOL_BINS): build/bin/%: tools/%
 	@mkdir -p build/bin
 	@cp -f $< $@
+	@sed -i '' '1s@/usr/bin/env python@$(PYTHON_BIN)@' $@
```
S-YOU: I am getting following error somehow
```
grumpy% make
sed: can't read 1s@/usr/bin/env python@/usr/bin/python@: No such file or directory
Makefile:98: recipe for target 'build/bin/grumpc' failed
make: *** [build/bin/grumpc] Error 2
```

S-YOU: Looks like `cp -f $< $@` is not done yet when sed try to read content of the file.
Not sure it is because of my xfs file system.

S-YOU: @trotterdylan, `sed -i''` without space fix the issue for me, btw

trotterdylan: I think 'sed -i' for in place edit is not supported on many systems. I'll come up with something different.

S-YOU: @trotterdylan, after some tries, on debian, sed -i'' without space only works and 
on OSX, sed -i '' with space only works for me.

S-YOU: http://stackoverflow.com/a/4247319/213464

S-YOU:
Make running ok on my debian and osx now.

PR #90, opened 2017-01-11
Revert "Update errno from go"

S-YOU:
Looks like importing directly from syscall throw errors when undefined on Mac OS X.
#89


trotterdylan:
Thanks for reverting. I think the issue is that not all errors are defined for all platforms. I still think the general approach is sound, though.

PR #91, opened 2017-01-11
Update .travis.yml to include OS X 

trotterdylan:


S-YOU:
I think we need this get merge.

trotterdylan:
There's still some issue with pip that I haven't figured out a solution to.

S-YOU:
Ah, right, I do get .functools-lru-cache warning/error at my home computer.

trotterdylan:
I see the .functools-lru-cache messages too but I think it's just a warning. The bigger issue is the pip error `error: must supply either home or prefix/exec-prefix -- not both`. You can see it in the latest travis run [here](https://travis-ci.org/google/grumpy/jobs/191536423).

S-YOU:
Some suggest to do `echo "[install]\nprefix=" > ~/.pydistutils.cfg`, but I am not sure about that.

trotterdylan:
I think I need to make some changes to the way we get and install pylint.

PR #93, opened 2017-01-12
Implement str.lower, upper, title

S-YOU:
Implemented str.lower, upper, title.
I haven't done capitalize, swapcase yet, because I can't figure equivalent one in go.


trotterdylan:
This is great. Thanks for pushing this! Merging now.

PR #95, opened 2017-01-12
add stdlib string

S-YOU:
This is from stdlib string, without _TemplateMetaclass, and its usage in Template class.

Also re.escape was added which is needed for string module.
(I moved lib/re.py to third_party/stdlib because code is from CPython)

Note that, this PR depends on PR #93, and #94. CI will fail without those.


third_party/stdlib/string.py
```diff
@@ -0,0 +1,658 @@
...
+#     )
+#     """
+
+#     def __init__(cls, name, bases, dct):
+#         super(_TemplateMetaclass, cls).__init__(name, bases, dct)
+#         if 'pattern' in dct:
+#             pattern = cls.pattern
+#         else:
+#             pattern = _TemplateMetaclass.pattern % {
+#                 'delim' : _re.escape(cls.delimiter),
```
trotterdylan: It looks like you've commented out all the parts that use the `re` module so maybe just remove that file and the import for now?

third_party/stdlib/re.py
```diff
@@ -0,0 +1,30 @@
+#
+# Secret Labs' Regular Expression Engine
```
trotterdylan: Stuff from pypy needs to be in a different subdir of third_party (maybe third_party/pypy).

S-YOU: I copied that from CPython @trotterdylan

third_party/stdlib/string.py
S-YOU: Ah, I commented that out because of Metaclass, super(...).init giving error.
Looks like re is no use there. let me know If I need to remove re from this PR.

S-YOU: removed that re module, and put it back to lib/ @trotterdylan

third_party/stdlib/re.py
trotterdylan: Oh! Cool okay. Thanks for the explanation. I think we'll probably want to get the code from PyPy in the end because it's pure Python so perhaps it's best if we do a separate PR that pulls in more of `re` from there.

third_party/stdlib/string.py
trotterdylan: metaclass and super should work, so if they're problematic it's a bug. But we can fix that in a separate PR.

trotterdylan:
This is great, thanks! Merging.

PR #97, opened 2017-01-12
Implement builtin map

S-YOU:
Implementing builtin map
Note that variable builtinMap is taken, so I am using builtinMapFn.


runtime/builtin_types.go
```diff
@@ -193,6 +193,26 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return Abs(f, args[0])
 }
 
+func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
```
trotterdylan: `map()` can take multiple sequences and then supply them all as args to the function:

```
map(...)
    map(function, sequence[, sequence, ...]) -> list

    Return a list of the results of applying the function to the items of
    the argument sequence(s).  If more than one sequence is given, the
    function is called with an argument list consisting of the corresponding
    item of each sequence, substituting None for missing values when not all
    sequences have the same length.  If the function is None, return a list of
    the items of the sequence (or a list of tuples if more than one sequence).
```

S-YOU: Oh, I will work again. Never knew that I can pass many sequence to map.

S-YOU:
Added support for multiple sequences, but I can't seem to figure two coverage on
```
if !raised.isInstance(StopIterationType) {
     return nil, raised
}
```
and 
```
// Replace nil with None
if len(list) > 1 {
	lastIndex := len(list) - 1
	for list[lastIndex] == nil {
		list[lastIndex] = None
		lastIndex--
	}
}
```

S-YOU:
Yeah, Indeed. I will move zip implemention as separate function.

S-YOU:
I am now using similar approach as zip implementation, but I can't able to find way to share code yet, because there is little different on innermost loop.

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		}
+		// Replace nil with None
+		if len(elems) > 1 {
+			lastIndex := len(elems) - 1
+			for elems[lastIndex] == nil {
+				elems[lastIndex] = None
+				lastIndex--
+			}
+		}
+		// NOTE: append does the capacity doubling logic as needed
```
trotterdylan: Remove

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		result = append(result, NewTuple(elems...).ToObject())
+	}
+	return result, nil
+}
+
+func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
```
trotterdylan: Call this `builtinMap` to match the pattern used by other builtin functions.

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return Abs(f, args[0])
 }
 
+func zipLongestImpl(f *Frame, args Args) ([]*Object, *BaseException) {
```
trotterdylan: Call this zipLongest since it's slightly simpler. Also, please move it to the end of the file. I try to group all the helper functions at the end of the file (although I'm sure I've missed some).

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	argc := len(args)
+	result := make([]*Object, 0, 2)
+	iters := make([]*Object, argc)
+	for i, arg := range args {
+		iter, raised := Iter(f, arg)
+		if raised != nil {
+			return nil, raised
+		}
+		iters[i] = iter
+	}
```
trotterdylan: Seems like you could factor this iterator collecting logic out of map() and zip() into a function that returns []*Object. But that could wait until a later PR if you prefer.

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+
+	for {
+		noItems := true
+		elems := make([]*Object, argc)
+		for i, iter := range iters {
+			elem, raised := Next(f, iter)
+			if raised != nil {
+				if raised.isInstance(StopIterationType) {
+					continue
```
trotterdylan: As in zip(), you need to RestoreExc(nil, nil)

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+				return nil, raised
+			}
+			noItems = false
+			elems[i] = elem
+		}
+		if noItems {
+			break
+		}
+		// Replace nil with None
+		if len(elems) > 1 {
```
trotterdylan: Could you set `elems[i] = None` when you detect StopIteration above?

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			return nil, raised
+		}
+		iters[i] = iter
+	}
+
+	for {
+		noItems := true
+		elems := make([]*Object, argc)
+		for i, iter := range iters {
+			elem, raised := Next(f, iter)
```
trotterdylan: Raising an exception requires allocating a new object which is relatively expensive. In general it's good to avoid as many allocations as possible. So rather than calling Next() on exhausted iterators, it would probably be cheaper to set `iters[i] = nil` and check for nil before calling Next().

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+
+	pred := toFunctionUnsafe(args[0])
+	result := make([]*Object, 0, 2)
+
+	z, raised := zipLongestImpl(f, args[1:])
+	if raised != nil {
+		return nil, raised
+	}
+	for _, tuple := range z {
+		if pred.Type() == NoneType {
```
trotterdylan: `if pred == None {`

runtime/builtin_types.go
```diff
@@ -193,6 +193,81 @@ func builtinAbs(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			if argc == 2 {
+				result = append(result, toTupleUnsafe(tuple).GetItem(0))
+			} else {
+				result = append(result, tuple)
+			}
+		} else {
+			ret, raised := pred.Call(f, toTupleUnsafe(tuple).elems, nil)
```
trotterdylan: Rather than allocating a Tuple and then not using it in most cases, maybe zipLongest should return `[][]*Object.`

runtime/builtin_types.go
S-YOU: Sure

runtime/builtin_types.go
S-YOU: Let me do this after both PR are merged, because switching between two branch to refactor or through web take time.

runtime/builtin_types.go
S-YOU: I cannot, builtinMap is this stuff
```
builtinMap := map[string]*Object{
		"__frame__":      newBuiltinFunction("__frame__", builtinFrame).ToObject(),
		"abs":            newBuiltinFunction("abs", builtinAbs).ToObject(),
		"all":            newBuiltinFunction("all", builtinAll).ToObject(),
```

S-YOU:
Fixed all, except renaming builtinMapFn to builtinMap, because builtinMap is in used by

```
builtinMap := map[string]*Object{
		"__frame__":      newBuiltinFunction("__frame__", builtinFrame).ToObject(),
		"abs":            newBuiltinFunction("abs", builtinAbs).ToObject(),
		"all":            newBuiltinFunction("all", builtinAll).ToObject(),
```

also added `initIters` function, to collect iterators, let me know if you have better name.

runtime/builtin_types.go
S-YOU: I thought nil where more than one, looks like that was not the case. 
Much simpler now. Thank you.

runtime/builtin_types.go
trotterdylan: Now that this is merged with the zip() change, would you mind updating builtinZip() to use initIters() as well?

S-YOU: Sure

S-YOU:
Rebased and updated to use initIters.

trotterdylan:
This is great! Thanks for doing this.

PR #98, opened 2017-01-12
export list.sort

S-YOU:
I think basic list.sort can be exported, most tests cases are already there.

Note that, cmd, key, and reversed options are not yet implemented.

### DIFF
--- a/runtime/list.go
+++ b/runtime/list.go
@@ -336,10 +336,20 @@ func listSetItem(f *Frame, o, key, value *Object) *BaseException {
 	return f.RaiseType(TypeErrorType, fmt.Sprintf("list indices must be integers, not %s", key.Type().Name()))
 }
 
+func listSort(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkMethodArgs(f, "sort", args, ListType); raised != nil {
+		return nil, raised
+	}
+	l := toListUnsafe(args[0])
+	l.Sort(f)
+	return None, nil
+}
+
 func initListType(dict map[string]*Object) {
 	dict["append"] = newBuiltinFunction("append", listAppend).ToObject()
 	dict["insert"] = newBuiltinFunction("insert", listInsert).ToObject()
 	dict["reverse"] = newBuiltinFunction("reverse", listReverse).ToObject()
+	dict["sort"] = newBuiltinFunction("sort", listSort).ToObject()
 	ListType.slots.Add = &binaryOpSlot{listAdd}
 	ListType.slots.Contains = &binaryOpSlot{listContains}
 	ListType.slots.Eq = &binaryOpSlot{listEq}
--- a/runtime/list.go
+++ b/runtime/list.go
@@ -337,6 +337,7 @@ func listSetItem(f *Frame, o, key, value *Object) *BaseException {
 }
 
 func listSort(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	// TODO: Support (cmp=None, key=None, reverse=False)
 	if raised := checkMethodArgs(f, "sort", args, ListType); raised != nil {
 		return nil, raised
 	}
--- a/runtime/list_test.go
+++ b/runtime/list_test.go
@@ -393,14 +393,10 @@ func TestListSetItem(t *testing.T) {
 
 func TestListSort(t *testing.T) {
 	fun := newBuiltinFunction("TestListSort", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
-		if raised := checkFunctionArgs(f, "TestListSort", args, ListType); raised != nil {
+		if _, raised := listSort(f, args, KWArgs{}); raised != nil {
 			return nil, raised
 		}
-		l := toListUnsafe(args[0])
-		if raised := l.Sort(f); raised != nil {
-			return nil, raised
-		}
-		return l.ToObject(), nil
+		return args[0], nil
 	}).ToObject()
 	cases := []invokeTestCase{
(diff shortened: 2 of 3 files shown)


runtime/list_test.go
```diff
@@ -408,6 +408,8 @@ func TestListSort(t *testing.T) {
 		{args: wrapArgs(newTestList(true, false)), want: newTestList(false, true).ToObject()},
 		{args: wrapArgs(newTestList(1, 2, 0, 3)), want: newTestRange(4).ToObject()},
 		{args: wrapArgs(newTestRange(100)), want: newTestRange(100).ToObject()},
+		{args: wrapArgs(1), wantExc: mustCreateException(TypeErrorType, "'TestListSort' requires a 'list' object but received a \"int\"")},
+		{args: wrapArgs(newTestList("foo", "bar"), 2), wantExc: mustCreateException(TypeErrorType, "'TestListSort' requires 1 arguments")},
```
meadori: Isn't this really testing the test and not the sort implementation?  Not sure that is necessary.

S-YOU: I think I did that because of coverage missing on 338~340.
The name 'TestListSort' is just because of fun is created inline there.

meadori: The first test that was added to exercise 338-340 is fine.  It is this test I am wondering about:

```
{args: wrapArgs(1), wantExc: mustCreateException(TypeErrorType, "'TestListSort' requires a 'list' object but received a \"int\"")},
```

Doesn't that just test that the `fun` testing function throws an exception? i.e. it doesn't test any implementation code, just test code.

S-YOU: Ah, you are right, I missed there is checkFunctionArgs inside that test function, removing last one...

S-YOU: Removed both, Thank you very much @meadori.

runtime/list_test.go
```diff
@@ -392,8 +392,8 @@ func TestListSetItem(t *testing.T) {
 }
 
 func TestListSort(t *testing.T) {
-	fun := newBuiltinFunction("TestListSetItem", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
-		if raised := checkFunctionArgs(f, "TestListSetItem", args, ListType); raised != nil {
+	fun := newBuiltinFunction("TestListSort", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
```
trotterdylan: To get some coverage over the listSort() code, you could make this test call list.sort() rather than *List.Sort(). Then you could test the arg validating logic as well.

runtime/list.go
```diff
@@ -334,10 +334,20 @@ func listSetItem(f *Frame, o, key, value *Object) *BaseException {
 	return f.RaiseType(TypeErrorType, fmt.Sprintf("list indices must be integers, not %s", key.Type().Name()))
 }
 
+func listSort(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
```
trotterdylan: As mentioned in the sorted() PR, list.sort() is a little more complicated than this:

`list.sort(cmp=None, key=None, reverse=False)`

A TODO is fine for now.

S-YOU:
Added
- python side tests
- TODO about cmp=None, key=None, reverse=False)
- and reworked testing code to pass through arguments

Please take a look, @trotterdylan, @meadori

runtime/list.go
S-YOU: Added a TODO

runtime/list_test.go
```diff
@@ -392,22 +392,20 @@ func TestListSetItem(t *testing.T) {
...
 
 func TestListSort(t *testing.T) {
-	fun := newBuiltinFunction("TestListSetItem", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
-		if raised := checkFunctionArgs(f, "TestListSetItem", args, ListType); raised != nil {
-			return nil, raised
-		}
-		l := toListUnsafe(args[0])
-		if raised := l.Sort(f); raised != nil {
+	fun := newBuiltinFunction("TestListSort", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		if _, raised := listSort(f, args, KWArgs{}); raised != nil {
```
trotterdylan: Prefer nil to KWArgs{}

trotterdylan: I prefer to get the list.sort attribute and then Call(). Using listSort() directly is kind of like a back door. You can fetch it once above `fun` and then use it on each test:

```
sort := mustNotRaise(GetAttr(NewRootFrame(), ListType.ToObject(), NewStr("sort"), nil))
fun := ... {
  if _, raised := sort.Call(f, args, nil); raised != nil {
  ...
```

S-YOU: fixed this.

S-YOU: Updated to reflect that. 

One question, so we are testing 
```
func (l *List) Sort(f *Frame) (raised *BaseException) {
```

Instead of 
```
func listSort(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
```

Correct?

trotterdylan: You're indirectly testing both. `GetAttr(f, ListType.ToObject(), NewStr("sort"), nil)` is the equivalent of `getattr(list, 'sort')`. By calling sort this way, you're invoking listSort, which is in turn calling `func (l *List) Sort()`.

trotterdylan:
This is great! Thanks for getting this done.

runtime/list_test.go
S-YOU: Thanks for the explanations, I got it now.

PR #100, opened 2017-01-12
Add builtin sorted

S-YOU:
Add builtin sorted, without cmp, key, and reverse options.


S-YOU:
Thanks @meadori, I didn't realize there is tests for python side.
--
Looks like I shouldn't include reversed in this PR. I didnt know about there is __reverse__ slot.

meadori:
Yeah, I would break the `reversed` one out into a separate PR.  It will probably be more involved.

S-YOU:
@meadori, Thank you, removed reversed

runtime/builtin_types.go
```diff
@@ -541,6 +541,18 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return s.ToObject(), nil
 }
 
+func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkFunctionArgs(f, "sorted", args, ObjectType); raised != nil {
+		return nil, raised
+	}
+	result, raised := listNew(f, ListType, Args{args[0]}, KWArgs{})
```
trotterdylan: `ListType.Call(f, Args{args[0]}, nil)`

runtime/builtin_types.go
```diff
@@ -541,6 +541,18 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return s.ToObject(), nil
 }
 
+func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkFunctionArgs(f, "sorted", args, ObjectType); raised != nil {
```
trotterdylan: `sorted()` is a little more complicated than this:

```
sorted(iterable, cmp=None, key=None, reverse=False) --> new sorted list
```

A TODO is fine. Then that functionality could be added to list.sort() and then passed when you call list.sort() below.

runtime/builtin_types_test.go
```diff
@@ -228,6 +228,13 @@ func TestBuiltinFuncs(t *testing.T) {
 		{f: "repr", args: wrapArgs(NewUnicode("abc")), want: NewStr("u'abc'").ToObject()},
 		{f: "repr", args: wrapArgs(newTestTuple("foo", "bar")), want: NewStr("('foo', 'bar')").ToObject()},
 		{f: "repr", args: wrapArgs("a", "b", "c"), wantExc: mustCreateException(TypeErrorType, "'repr' requires 1 arguments")},
+		{f: "sorted", args: wrapArgs(NewList()), want: NewList().ToObject()},
+		{f: "sorted", args: wrapArgs(newTestList("foo", "bar")), want: newTestList("bar", "foo").ToObject()},
```
trotterdylan: Maybe some test cases that are other iterable types like dict and tuple.

runtime/builtin_types.go
meadori: I noticed that as well, but didn't explicitly mention it b/c @S-YOU put `Add builtin sorted, without cmp, key, and reverse options.` in the change request message.

S-YOU:
Updated following
- TODO about (cmp=None, key=None, reverse=False)
- Use ListType.Call instead of listNew
- Added tests for Tuple and Dict
- Added some python side testing

Please take a look, @meadori, @trotterdylan

runtime/builtin_types.go
```diff
@@ -541,6 +541,19 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return s.ToObject(), nil
 }
 
+func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	//TODO: Support (cmp=None, key=None, reverse=False)
```
trotterdylan: Space after `//` in comments:

`// TODO: ...`

S-YOU:
@trotterdylan, added space after comment //, thank you very much.

PR #102, opened 2017-01-12
Implement builtin zip

S-YOU:
Let me know if my approach is not right.
More test case has to be done. Will do it my time tomorrow.

### DIFF
--- a/runtime/builtin_types.go
+++ b/runtime/builtin_types.go
@@ -589,11 +589,11 @@ Outer:
 				if raised.isInstance(StopIterationType) {
 					break Outer
 				}
+				f.RestoreExc(nil, nil)
 				return nil, raised
 			}
 			elems[i] = elem
 		}
-		// NOTE: append does the capacity doubling logic as needed
 		result = append(result, NewTuple(elems...).ToObject())
 	}
 	return NewList(result...).ToObject(), nil
--- a/runtime/builtin_types.go
+++ b/runtime/builtin_types.go
@@ -16,7 +16,6 @@ package grumpy
 
 import (
 	"fmt"
-	"math"
 	"math/big"
 	"os"
 	"unicode"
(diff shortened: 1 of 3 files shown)


runtime/builtin_types.go
```diff
@@ -552,6 +553,37 @@ func builtinUniChr(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return NewUnicodeFromRunes([]rune{rune(i)}).ToObject(), nil
 }
 
+func builtinZip(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	seqs := [][]*Object{}
+	lenHint := math.MaxInt64
+	for _, arg := range args {
+		if raised := checkFunctionArgs(f, "zip", Args{arg}, ObjectType); raised != nil {
```
trotterdylan: This check will never fail so just remove it.

runtime/builtin_types.go
```diff
@@ -552,6 +553,37 @@ func builtinUniChr(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			return nil, raised
+		}
+		seq, raised := seqNew(f, Args{arg})
```
trotterdylan: It's expensive to allocate all these slices, so I would recommend creating an iterator for each arg and calling next over all the iterators until one is exhausted.

That means you won't get a lenHint which makes it a little more expensive to build the result list, but you can grow the list slice by doubling its capacity every time to minimize the number of reallocations.

runtime/builtin_types.go
```diff
@@ -552,6 +553,37 @@ func builtinUniChr(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			lenHint = len(seq)
+		}
+	}
+	list := make([]*Object, lenHint)
+	for i := 0; i < lenHint; i++ {
+		tuple := make([]*Object, len(args))
+		for j := 0; j < len(args); j++ {
+			tuple[j] = seqs[j][i]
+		}
+		t := toTupleUnsafe(newObject(TupleType))
```
trotterdylan: `t := NewTuple(tuple...)`

runtime/builtin_types.go
```diff
@@ -552,6 +553,37 @@ func builtinUniChr(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		t.elems = tuple
+		list[i] = t.ToObject()
+	}
+	l := toListUnsafe(newObject(ListType))
```
trotterdylan: `l := NewList(list...)`

S-YOU:
@trotterdylan, rewrote most part to reflect iter. PTAL.
I didnt manually make slice grow, but just using append.

S-YOU:
@trotterdylan, updated code to use slice growth by double, by using 1 dimensional slice.
Please take a look.

S-YOU:
Thanks for the implementation @trotterdylan. It just works :+1:

runtime/builtin_types.go
```diff
@@ -565,6 +565,40 @@ func builtinUniChr(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			elem, raised := Next(f, iter)
+			if raised != nil {
+				if raised.isInstance(StopIterationType) {
+					break Outer
+				}
+				return nil, raised
+			}
+			elems[i] = elem
+		}
+		// NOTE: append does the capacity doubling logic as needed
```
trotterdylan: Remove this comment.

runtime/builtin_types.go
```diff
@@ -565,6 +565,40 @@ func builtinUniChr(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	}
+
+Outer:
+	for {
+		elems := make([]*Object, argc)
+		for i, iter := range iters {
+			elem, raised := Next(f, iter)
+			if raised != nil {
+				if raised.isInstance(StopIterationType) {
+					break Outer
```
trotterdylan: You'll also need to clear the exception state on the frame by calling `f.RestoreExc(nil, nil)` similar to [here](https://github.com/google/grumpy/blob/master/runtime/core.go#L122).

S-YOU: Thank you for explanation, I noticed that line somewhere, but didn't understand.

runtime/builtin_types.go
S-YOU: Done

trotterdylan:
Thanks for implementing this! Merging now.

PR #103, opened 2017-01-12
Make `list.sort` stable

meadori:
Python gaurantees that its builtin sort for mutable sequences is stable.
For more details, see note (9) in:

  * https://docs.python.org/2/library/stdtypes.html#mutable-sequence-types

I couldn't come up with a decent way to test this.


meadori:
Yeah, I thought about that.  I looked at both the CPython and Go standard library test suites.  They both did something similar, but I questioned the value too.

Anyway, thanks for merging!

PR #105, opened 2017-01-13
Add _struct, _md5 to third_party/pypy

S-YOU:
Added _struct, _md5, md5(wrapper to _md5) to third_party/pypy.

Hard-coded `byteorder = 'little'` on sys module for now, which is needed by _struct
Let me know if I better hardcode inside _struct module.

Note that I cannot add wrapper struct.py because struct is a keyword in go as you know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,8 +54,8 @@ RUNNER = $(RUNNER_BIN) $(COMPILER) $(RUNTIME) $(STDLIB)
 
 GRUMPY_STDLIB_SRCS := $(shell find lib -name '*.py')
 GRUMPY_STDLIB_PACKAGES := $(foreach x,$(GRUMPY_STDLIB_SRCS),$(patsubst lib/%.py,%,$(patsubst lib/%/__init__.py,%,$(x))))
-THIRD_PARTY_STDLIB_SRCS := $(wildcard third_party/stdlib/*.py)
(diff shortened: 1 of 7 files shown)


trotterdylan:
Awesome! Do we need a struct.py file as well?

S-YOU:
@trotterdylan 
How? `package struct` throw error in golang. it is a keyword, you know.

trotterdylan:
Oh, heh. Yeah I forgot about that. Then I think we'll need some special handling in grumpc for modules that have a Go keyword as their name. Are you thinking a leading prefix like in `_struct` is sufficient?

S-YOU:
Both struct, and _struct are existing python module, we might need like πg prefix to module too, but I am worried that that would be too complicated, also on the file system.

trotterdylan:
Currently it's OK to have a struct.py because it will be aliased as π_grumpyΓlibΓstruct in the Go code. The only trouble is if somebody imports it in hand-written Go code they'll have to alias it like:

`import _struct "grumpy/lib/struct"`

But that seems OK too.

S-YOU:
Using the library is ok with π_grumpyΓlibΓstruct, but the problem is you cannot compile struct module, because in build/src/grumpy/lib/_struct/module.go first line get generated is

package struct

which fail to compile the grumpy itself.

trotterdylan:
Ah! Good point. Then it seems like the options are:

1. Always prefix Grumpy package names so that there is never a keyword collision
2. Have special handling in grumpc that changes the way imports work for modules that are named the same as a Go keyword.
3. ???

PR #107, opened 2017-01-13
Add str interpolation for %x

S-YOU:
Added str interpolation for %x.


S-YOU:
Nope, I will do that way.

S-YOU:
Rebased, Please take a look.

runtime/str.go
```diff
@@ -625,11 +625,20 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
...
-			if raised != nil {
-				return nil, raised
+			if matches[7] == "d" {
+				s, raised := ToStr(f, i)
+				if raised != nil {
+					return nil, raised
+				}
+				buf.WriteString(s.Value())
+			}
+			if matches[7] == "x" || matches[7] == "X" {
```
trotterdylan: This can just be an `} else {`

runtime/str.go
```diff
@@ -625,11 +625,20 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
...
 			}
-			s, raised := ToStr(f, i)
-			if raised != nil {
-				return nil, raised
+			if matches[7] == "d" {
+				s, raised := ToStr(f, i)
+				if raised != nil {
+					return nil, raised
+				}
+				buf.WriteString(s.Value())
```
trotterdylan: Would be nice to have a single buf.WriteString() call to accompany valueIndex++. Maybe declare a `var val string` and then populate it in each conditional branch.

S-YOU: declaring var is on the top of func or inside case is ok?

runtime/str.go
S-YOU: Ok, sure.

runtime/str.go
trotterdylan: Inside the case is ok

trotterdylan:
Looks good! Thanks for doing this.

PR #109, opened 2017-01-14
Try `__cmp__` when rich comparisons aren't handy

meadori:
When comparing two objects and the appropriate rich comparison
method doesn't exist, the `__cmp__` method should be tried. For
example, if `a == b` is executed and neither `a` or `b` have
the appropriate rich comparison methods, then the equivalent
of `a.__cmp__(b) == 0` should be tried.

### DIFF
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -893,7 +893,59 @@ func (op compareOp) slot(t *Type) *binaryOpSlot {
 }
 
 func compareRich(f *Frame, op compareOp, v, w *Object) (*Object, *BaseException) {
-	// TODO: Support __cmp__.
+	r, raised := tryRichCompare(f, op, v, w)
+	if raised != nil {
+		return nil, raised
+	}
+	if r != NotImplemented {
+		return r, nil
+	}
+	return try3wayToRichCompare(f, op, v, w)
+}
+
+// convert3wayToObject converts the integer results from a 3-way
+// comparison to a suitable boolean value for the given rich
+// comparison op.
+func convert3wayToObject(op compareOp, c int) *Object {
+	b := false
+	switch op {
+	case compareOpLT:
+		b = c < 0
+	case compareOpLE:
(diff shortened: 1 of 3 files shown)


runtime/core.go
```diff
@@ -893,7 +893,62 @@ func (op compareOp) slot(t *Type) *binaryOpSlot {
...
+	case compareOpEq:
+		b = c == 0
+	case compareOpNE:
+		b = c != 0
+	case compareOpGE:
+		b = c >= 0
+	case compareOpGT:
+		b = c > 0
+	}
+	if b {
```
trotterdylan: `return GetBool(b).ToObject()`

meadori: Much better.  Thanks!

meadori: I always use `make cover` to check for coverage results, but that seems to give inconsistent results.  I will try the command you posted above.  Thanks.

trotterdylan: make cover cross references with the lines changed according to git diff. That workflow doesn't work very well since there are typically multiple commits per PR. You can run `make DIFF_COMMIT=xyz cover` to have it cross reference with a different commit but it's a bit awkward. I need to find a nicer way to get coverage into a normal workflow.

S-YOU: Nice to know DIFF_COMMIT, so I guess `make DIFF_COMMIT=master cover` could works against master, right?

meadori:
I addressed the one code change and added plenty of tests.  We should be at full coverage for the 3-way and rich comparison implementations now.

trotterdylan:
This looks really good. Thanks for the thorough tests! Merging.

runtime/core.go
trotterdylan: Isn't master just head in a forked repo? I think what you might want upstream/master or something but I'm not sure that works.

S-YOU: Ah, you are right.

PR #110, opened 2017-01-14
add lstrip, rstrip

S-YOU:
Added lstrip, rstrip based on existing strip
Note that, python side tests still have to be done.


runtime/str.go
```diff
@@ -489,26 +502,30 @@ func strStrip(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	byteSlice := []byte(s.Value())
 	numBytes := len(byteSlice)
 	lindex := 0
-LeftStrip:
-	for ; lindex < numBytes; lindex++ {
-		b := byteSlice[lindex]
-		for _, c := range chars {
-			if b == c {
-				continue LeftStrip
+	if sideFlag&1 == 1 {
```
trotterdylan: Since the zero flag value doesn't make much sense, I would make this an enum. And make it typed so that the values passed as parameters are always valid:

```
type stripSide int

const (
  stripSideLeft stripSide = iota
  stripSideRight stripSide
  stripSideBoth stripSide
)

func strStripImpl(..., side stripSide) ... {
```

S-YOU:
@trotterdylan , thank you very much, updated those.

PR #111, opened 2017-01-14
Add binascii module to third_party/pypy

S-YOU:
Added binascii module third_party/pypy
Tests are also borrowed from pypy, put it in testing/ folder, let me know if I should put somewhere else.

Few tests are failing, I have put some TODOs instead.


S-YOU:
Removed the test file from this PR.
I tried to figure out addition to Makefile, but no luck.

PR #113, opened 2017-01-14
[TMP] add str.find, str.index, str.zfill and a fix

S-YOU:
- Implemented str.find, str.index and str.zfill
- No unicode support yet
- Fixed adjustIndex and test case, `"".startswith("",1)` and `"".endswith("",1)` returning True, 
which should be False
- added sys.maxsize = MaxInt64 - 1 (`__go__.math`)
- I am borrowing some tests from CPython


lib/sys.py
```diff
@@ -17,13 +17,15 @@
...
 from __go__.runtime import Version
+from __go__.math import MaxInt64
 
 argv = []
 for arg in Args:
   argv.append(arg)
 
 goversion = Version()
 maxint = MaxInt
+maxsize = MaxInt64 - 1
```
trotterdylan: On 32 bit systems, this should be MaxInt32, right?

runtime/str.go
```diff
@@ -205,6 +205,36 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return strCompare(v, w, False, True, False), nil
 }
 
+func strFind(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	expectedTypes := []*Type{StrType, StrType, IntType, IntType}
```
trotterdylan: If args[1] is unicode then the CPython implementation dispatches to unicode.find(): https://github.com/python/cpython/blob/2.7/Objects/stringobject.c#L1724

If you'd rather not add unicode.find() to this PR, then please add a TODO.

runtime/str.go
```diff
@@ -264,6 +294,17 @@ func strHash(f *Frame, o *Object) (*Object, *BaseException) {
 	return h.ToObject(), nil
 }
 
+func strIndex(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	index, raised := strFind(f, args, nil)
```
trotterdylan: If checkMethodArgs() raises an exception then it will say `'find' of 'str' requires ...` rather than `'index' of 'str' ...`

CPython gets around this by giving a really generic error message: `TypeError: find/rfind/index/rindex() takes at least 1 argument (0 given)`

runtime/str.go
```diff
@@ -711,8 +755,6 @@ func adjustIndex(i, l int) int {
 		i = 0
 	case i < 0:
 		i += l
-	case i > l:
-		i = l
```
trotterdylan: I think this check is crucial. If we don't adjust i in this case and we use it to slice a string, Go will panic: `panic: runtime error: slice bounds out of range`. The intended postcondition of this function is that i is valid for slicing a string of length l.

Is there another way to fix the bug you found? Perhaps we need to remove this function and do the logic inline so that we can handle that case?

testing/str_test.py
```diff
@@ -12,7 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# pylint: disable=redefined-outer-name
+# pylint: disable=redefined-outer-name,W0703
```
trotterdylan: Does W0703 have a descriptive name we could use here instead?

testing/str_test.py
```diff
@@ -12,7 +12,20 @@
...
+
+import sys
+
+
+def checkequal(expected, s, attr, *args):
```
trotterdylan: Seems like this method doesn't make the tests any more readable than just writing out the assert statement. Compare `checkequal('123', '123', 'zfill', 2)` and `assert '123'.zfill(2) == '123'`. The latter is shorter is much more difficult to misread. I'd suggest just writing out the asserts.

testing/str_test.py
```diff
@@ -12,7 +12,20 @@
...
+  assert getattr(s, attr)(*args) == expected
+
+
+def checkraises(expected, s, attr, *args):
```
trotterdylan: _checkraises() since this is a private helper function.

Also, I'd recommend matching the semantics of assertRaises as closely as possible in the unittest library since that may be familiar to people reading the code.

lib/sys.py
S-YOU: I dont see how to get platform specific value from go.
It is currently using on a minor test, pull from CPython, 
I can use maxint, not much different I think.
Let me know If you want me to remove it.

testing/str_test.py
S-YOU: Sure, I will update. I copied those codes from CPython tests, I was lazy to rewrite it.

testing/str_test.py
S-YOU: Sure

runtime/str.go
S-YOU: I will add TODO for now

runtime/str.go
S-YOU: I was doing bound check on outside of that function, so I didn't notice that check was needed.

That case block has side effects,`"".startswith("", 1)` return True, but it should be False, when you run in real python.

testing/str_test.py
S-YOU: It was for this clause, but may be I should do that way in first place. Will remove.
`except Exception as e`

lib/sys.py
trotterdylan: I think generally we use grumpy.MaxInt as a max size for sequences in Grumpy. So yeah I think you can just make maxint and maxsize the same.

runtime/str.go
trotterdylan: Maybe the adjustIndex helper function isn't very useful and we should just do the logic inline? I'm not really sure.

S-YOU: It is usefull for adjusting negative values like [-1:], but not for bound checking I guess.
I am now adding this bound checkout before slicing.
```
       if start > end || start > l {
		return NewInt(-1).ToObject(), nil
	}
```

runtime/str.go
S-YOU: I named it find/index for now.

S-YOU:
Rebased, bound checking logic outside adjustIndex is already done.
Also adjustIndex currently in used by startswith, endswith, find, and index.
Current tests are passed, I am still trying to find out other possible slice out of bound errors.

runtime/str.go
```diff
@@ -796,6 +839,28 @@ func strUpper(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	}
+	s := toStrUnsafe(args[0]).Value()
+	l := len(s)
+	width := toIntUnsafe(args[1]).Value()
+	if width <= l {
+		return args[0], nil
+	}
+	buf := bytes.Buffer{}
+	buf.Grow(width)
+	if len(s) > 0 && (s[0] == '-' || s[0] == '+') {
```
trotterdylan: You've already calculated `l := len(s)` so use `l` here instead.

runtime/str.go
```diff
@@ -205,6 +205,37 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+		start = adjustIndex(toIntUnsafe(args[2]).Value(), l)
+	}
+	if argc == 4 {
+		end = adjustIndex(toIntUnsafe(args[3]).Value(), l)
+	}
+	if start > end || start > l {
+		return NewInt(-1).ToObject(), nil
+	}
+	sep := toStrUnsafe(args[1]).Value()
+	s = s[start:end]
```
trotterdylan: If the user passes end > len(s), I think this will panic. The nice thing about clamping start and end to 0 <= start/end <= len(s) is that it's guaranteed to be safe to pass into slice bounds. You need to be able to determine `start > l` so I would suggest that you store `toIntUnsafe(args[2]).Value()` as a separate value (e.g. startRaw) and check `startRaw > l` to address the bug you found, but then calculate `start = adjustIndex(startRaw, l)`

runtime/str.go
```diff
@@ -205,6 +205,37 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return strCompare(v, w, False, True, False), nil
 }
 
+func strFind(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	// TODO: Support unicode sub string
```

runtime/str.go
```diff
@@ -205,6 +205,37 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+		expectedTypes = expectedTypes[:argc]
+	}
+	if raised := checkMethodArgs(f, "find/index", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	s := toStrUnsafe(args[0]).Value()
+	l := len(s)
+	start, end := 0, l
+	if argc >= 3 {
+		start = adjustIndex(toIntUnsafe(args[2]).Value(), l)
```
S-YOU: Ok, will add a TODO. I need to learn a bit more about using slots.
Didn't know about IndexInt, thank you.

runtime/str.go

runtime/str.go
S-YOU: Missed that, thank you

S-YOU:
- Rewrote adjustIndex, with parameter `upperBound bool`, because it turns out that 
```
case i > l:
    i = l
```
check was necessary for `end`, but not for `start`

- Also now using IndexInt, which support __index__ slot
- Added nil check inside `IndexInt`, because `Index` may return nil, nil

trotterdylan:
Sorry to suggest this late into the review, but would you mind breaking this into multiple PRs for index, find and zfill? I think it'd make it easier to reason about.

S-YOU:
Sure, I will work on it.

S-YOU:
Closing this, will submit str.index after str.find is ok.

PR #116, opened 2017-01-14
Implement `str.capitalize`

meadori:


runtime/str_test.go
```diff
@@ -253,6 +253,10 @@ func TestStrMethods(t *testing.T) {
 		want       *Object
 		wantExc    *BaseException
 	}{
+		{"capitalize", wrapArgs(""), NewStr("").ToObject(), nil},
+		{"capitalize", wrapArgs("foobar"), NewStr("Foobar").ToObject(), nil},
+		{"capitalize", wrapArgs("FOOBAR"), NewStr("Foobar").ToObject(), nil},
+		{"capitalize", wrapArgs("foobar", 123), nil, mustCreateException(TypeErrorType, "'capitalize' of 'str' requires 1 arguments")},
```
trotterdylan: Include some multi-byte characters, e.g. "ù"

runtime/str.go
```diff
@@ -157,6 +157,23 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return NewStr(stringV + stringW).ToObject(), nil
 }
 
+func strCapitalize(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	expectedTypes := []*Type{StrType}
```
meadori: Whoops, copypasta on my part.  Will fix.

runtime/str.go
```diff
@@ -157,6 +157,23 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	if raised := checkMethodArgs(f, "capitalize", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	s := toStrUnsafe(args[0]).Value()
+	buf := bytes.Buffer{}
+	numBytes := len(s)
+	if numBytes > 0 {
+		buf.WriteRune(unicode.ToUpper(rune(s[0])))
```
meadori: Given that Go does support the same notion of locales, then I don't see much we can do for now other than list it as a known limitation somewhere.  Maybe things will change in the future, but I think a partial implementation with a documented limitation is a good start (better than not implementing the method at all).

I will look into limiting the conversions to `0 <= b < 128`.  Thanks!

runtime/str_test.go
meadori: Will do.

meadori:
The latest version should address the review feedback.

README.md
```diff
@@ -29,6 +29,10 @@ supported).
    would add significant overhead when calling into and out of extension
    modules.
 
+3. C locale support: Go doesn't support locales in the same way that C does. As such,
```
trotterdylan: I think this is more like something that isn't *yet* supported. I suspect we'll need to support some variation on locales.

runtime/str.go
```diff
@@ -157,6 +157,29 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	writeRune := func(r rune, transform func(r rune) rune) {
+		if r <= unicode.MaxASCII {
+			r = transform(r)
+		}
+		buf.WriteRune(r)
+	}
+	if numRunes > 0 {
+		writeRune(runes[0], func(r rune) rune { return unicode.ToUpper(r) })
+	}
+	for i := 1; i < numRunes; i++ {
```
trotterdylan: For str I think you want to iterate over bytes, not runes. IIUC []rune(s) will treat s as a UTF-8 encoded string and convert it to 32-bit runes. So multi-byte characters will be treated as single runes and capitalized. This is not the behavior in CPython which seems to just capitalize with respect to ASCII:

```
>>> locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
'en_US.UTF-8'
>>> print u'вол'.capitalize()
Вол
>>> u'вол'.encode('utf8')
'\xd0\xb2\xd0\xbe\xd0\xbb'
>>> '\xd0\xb2\xd0\xbe\xd0\xbb'.capitalize()
'\xd0\xb2\xd0\xbe\xd0\xbb'```

Comment (meadori) [inline runtime/str.go]: I think my current implementation does the right thing for this case:

```
$ cat foo.py 
print '\xd0\xb2\xd0\xbe\xd0\xbb'
print '\xd0\xb2\xd0\xbe\xd0\xbb'.capitalize()
$ ./build/bin/grumpc foo.py > foo.go
$ go run foo.go                     
вол
вол
$ python foo.py 
вол
вол
```

I ran into problems iterating over bytes before because for the multi-byte character case each byte ended up being passed to `buf.WriteRune` separately, which caused issues (presumably b/c the byte was treated as a rune and encoded which in turn caused more bytes to appear in the output).

I could be missing something, though.  My Unicode-fu is a little weak :smile:

File: runtime/str.go [hunk 6]
```diff
@@ -157,6 +157,29 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	buf := bytes.Buffer{}
+	runes := []rune(s)
+	numRunes := len(runes)
+	writeRune := func(r rune, transform func(r rune) rune) {
+		if r <= unicode.MaxASCII {
+			r = transform(r)
+		}
+		buf.WriteRune(r)
+	}
+	if numRunes > 0 {
```
trotterdylan: I would handle the empty string case specially by exiting early and then just doing the rest of the logic unconditionally.

runtime/str.go
```diff
@@ -157,6 +157,29 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
...
 
+func strCapitalize(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	if raised := checkMethodArgs(f, "capitalize", args, StrType); raised != nil {
+		return nil, raised
+	}
+	s := toStrUnsafe(args[0]).Value()
+	buf := bytes.Buffer{}
+	runes := []rune(s)
+	numRunes := len(runes)
+	writeRune := func(r rune, transform func(r rune) rune) {
```
trotterdylan: I find the inline function a little confusing , compounded by the fact that it takes a callback. I would do this logic inline instead. It's not that much repetition.

runtime/str.go
trotterdylan: OK, I think I understand. Because you're only modifying chars < 128, `в` (U+1074) isn't affected. If a multibyte character contained bytes < 128 then CPython would do something differe, but I *think* that's never true in UTF-8. So you're right, in fact, this code should work but kind of by coincidence. It also seems a bit roundabout considering all we're interested in converting are the ASCII chars.

IMO the simplest and most efficient way to implement this are two tables for mapping lowercase ASCII chars to uppercase and vice versa. But you could also do it by checking `s[i] >= 'A' && s[i] <= 'Z'` and then setting `s[i] = s[i] - 'A' + 'a'` to go lowercase (and similar for uppercase).

meadori: Hmmm, after rereading what I have is way too complicated.  How about something like the following? :

```
s := toStrUnsafe(args[0])
sv := s.Value()
if len(sv) == 0 {
        return s.ToObject(), nil
}
return NewStr(strings.ToUpper(sv[0:1]) + strings.ToLower(sv[1:])).ToObject(), nil
```

It is more in line with `str.lower`, `str.upper`, etc...

trotterdylan: I like the simplicity of this and agree something like this would be best. But again, I think the trouble is that the Go standard library will affect UTF-8 characters different than Python does for str. In Python, str is just bytes and it seems like it only affects ASCII characters. The ToLower() part of the above will cause multi-byte characters like the last two in "ВОЛ" to switch to their lowercase equivalents, which wouldn't happen for a UTF-8 encoded equivalent str in CPython.

meadori: I see.  Thank you for talking this through with me.

FWIW, I think we have a similar problem with `str.title`:

```
$ cat foo.py 
print '\xd0\xb2\xd0\xbe\xd0\xbb'.title()
$ python foo.py                     
вол
$ ./build/bin/grumpc foo.py > foo.go
$ go run foo.go                     
Вол
```

trotterdylan: Yeah I think you're right :\

meadori:
Let's try this again :smile:  Thanks for all the review iterations @trotterdylan.  Much appreciated.

runtime/str.go
```diff
@@ -157,6 +157,32 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	}
+	s := toStrUnsafe(args[0])
+	sv := s.Value()
+	numBytes := len(sv)
+	if numBytes == 0 {
+		return s.ToObject(), nil
+	}
+	buf := bytes.Buffer{}
+	if sv[0] >= 'a' && sv[0] <= 'z' {
+		buf.WriteByte(sv[0] - 'a' + 'A')
```
trotterdylan: You could store a constant with value `'A' - 'a'` and then add and subtract that as necessary.

runtime/str.go
```diff
@@ -157,6 +157,32 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	if raised := checkMethodArgs(f, "capitalize", args, StrType); raised != nil {
+		return nil, raised
+	}
+	s := toStrUnsafe(args[0])
+	sv := s.Value()
+	numBytes := len(sv)
+	if numBytes == 0 {
+		return s.ToObject(), nil
+	}
+	buf := bytes.Buffer{}
```
trotterdylan: `bytes.Buffer{}` might be overkill here. You could preallocate a []byte of the correct size and write to that directly.

meadori: Fixed.

runtime/str.go
meadori: Fixed.

PR #117, opened 2017-01-15
Implement list.pop

S-YOU:


runtime/list.go
```diff
@@ -291,6 +300,29 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+		i = toIntUnsafe(args[1]).Value()
+	}
+	l := toListUnsafe(args[0])
+	if i < 0 {
+		i += len(l.elems)
+	}
+	if i >= len(l.elems) || i < 0 {
+		return nil, f.RaiseType(IndexErrorType, "list index out of range")
+	}
+	return l.Pop(i), nil
```
trotterdylan: You need to wrap the length validation in the mutex, since otherwise concurrent calls on the same object could race. I suggest removing the `(*List) Pop` method and just do the whole mutation in listPop.

runtime/list_test.go
```diff
@@ -353,6 +353,36 @@ func TestListIteratorIter(t *testing.T) {
...
+		}
+		return args[0], nil
+	}).ToObject()
+	cases := []invokeTestCase{
+		{args: wrapArgs(newTestList(-1, 0, 1)), kwargs: wrapKWArgs("result", True), want: NewInt(1).ToObject()},
+		{args: wrapArgs(newTestList(-1, 0, 1)), want: newTestList(-1, 0).ToObject()},
+		{args: wrapArgs(newTestList(-1, 0), 0), kwargs: wrapKWArgs("result", True), want: NewInt(-1).ToObject()},
+		{args: wrapArgs(newTestList(-1, 0, 1), 0), want: newTestList(0, 1).ToObject()},
+		{args: wrapArgs(newTestList(-1, 0, 1), None), kwargs: wrapKWArgs("result", True), wantExc: mustCreateException(TypeErrorType, "'pop' requires a 'int' object but received a 'NoneType'")},
+		{args: wrapArgs(newTestList(-1, 0, 1), None), wantExc: mustCreateException(TypeErrorType, "'pop' requires a 'int' object but received a 'NoneType'")},
```
trotterdylan: Test the IndexError case

runtime/list.go
```diff
@@ -291,6 +300,29 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	expectedTypes := []*Type{ListType, IntType}
+	if argc == 1 {
+		expectedTypes = expectedTypes[:1]
+	}
+	if raised := checkMethodArgs(f, "pop", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	i := -1
+	if argc == 2 {
+		i = toIntUnsafe(args[1]).Value()
```
S-YOU: I actually tried that on CPython. It didn't work.
```
>>> class A(object):
...   def __index__(self):
...     return 1
... 
>>> [1,2,3].pop(A())
TypeError: an integer is required
```
Is it still ok to add it?

runtime/list.go
S-YOU: Ok, sure

runtime/list.go
trotterdylan: Oh! Good catch. I just assumed. If that's the way it works in CPython then it's fine to leave it out. However, you still need to support `long`:

```
>>> [1,2,3].pop(2L)
3
```

runtime/list_test.go
S-YOU: Sure. I missed that.

runtime/list.go
S-YOU: Partially supporting long with `__index__` mean I should not use IndexInt? or Doesn't matter if with `__index__` support, even if CPython not support it?

S-YOU:
Updated with following
- inline listPop without Pop method on and wrap with mutex.Lock
- using IndexInt now, let me know if that is not ok
- This PR also included nil check on IndexInt in core.go becuase it's panic, 
there is another one on str.find PR too, so probably may cause small merge conflict later on.

runtime/list.go
trotterdylan: It's actually a bit more complicated than that :\

```
>>> class A(object):
...  def __int__(self):
...   return 1
...
>>> [1,2,3].pop(A())
2
```

I think the easiest thing to do is something like:

```
o, raised := IntType.Call(f, Args{args[1]}, nil)
if raised != nil {
  return nil, raised
}
i = toIntUnsafe(o).Value()
```

That will handle int and long properly as well as types that define `__int__`.

S-YOU: Oh, ok. Thanks for that.

S-YOU:
Updated to use IntType.Call. 

One other issue about mutex.Lock
I got deadlock with this code
```
	l.mutex.Lock()
        if i >= len(l.elems) || i < 0 {
		return nil, f.RaiseType(IndexErrorType, "list index out of range")
	}
	item := l.elems[i]
	l.elems = append(l.elems[:i], l.elems[i+1:]...)
	l.mutex.Unlock()
```

But not with this
```
        if i >= len(l.elems) || i < 0 {
		return nil, f.RaiseType(IndexErrorType, "list index out of range")
	}
	l.mutex.Lock()
	item := l.elems[i]
	l.elems = append(l.elems[:i], l.elems[i+1:]...)
	l.mutex.Unlock()
```

Any ideas?

runtime/list.go
```diff
@@ -291,6 +291,37 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return listCompare(f, toListUnsafe(v), w, NE)
 }
 
+func listPop(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
```
trotterdylan: `kwargs` is unused so call it `_`

runtime/list.go
```diff
@@ -291,6 +291,37 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+		i = toIntUnsafe(o).Value()
+	}
+	l := toListUnsafe(args[0])
+	if i < 0 {
+		i += len(l.elems)
+	}
+	if i >= len(l.elems) || i < 0 {
+		return nil, f.RaiseType(IndexErrorType, "list index out of range")
+	}
+	l.mutex.Lock()
```
trotterdylan: You must unlock before returning. That's why you were getting a deadlock with the other code. But this code is still racy. The whole operation including checking the length needs to be atomic, something like:

```
l.mutex.Lock()
numElems := len(l.elems)
if i < 0 {
  i += numElems
}
var item *Object
var raised *BaseException
if i >= numElems || i < 0 {
  raised = f.RaiseType(IndexErrorType, "list index out of range")
} else {
  item = l.elems[i]
  l.elems = append(l.elems[:i], l.elems[i+1:]...)
}
l.mutex.Unlock()
return item, raised
```

runtime/list_test.go
```diff
@@ -353,6 +353,39 @@ func TestListIteratorIter(t *testing.T) {
 	}
 }
 
+func TestListPop(t *testing.T) {
+	fun := newBuiltinFunction("TestListPop", func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
```
trotterdylan: I would get rid of the kwargs here and have the function return (pop result, list). Something like:

```
pop := mustNotRaise(GetAttr(NewRootFrame(), ListType.ToObject(), NewStr("pop"), nil))
fun := wrapFuncForTest(func(f *Frame, l *List, args ...*Object) (*Tuple, *BaseException) {
  result, raised := pop.Call(f, wrapArgs(l, args...), nil)
  if raised != nil {
    return nil, raised
  }
  return newTestTuple(result, l), nil
})
```

Sorry for wrapFuncForTest() being poorly documented. It's pretty smart about checking that arguments are of the right type and variadic arguments and such so it's a nice way to avoid having to do a lot of validation inside a `func(*Frame, Args, KWArgs)`.

S-YOU: Looks like I cannot pass `l` and `args...` together to wrapArgs. 
Should I create new slice there inside the function?

runtime/list.go
S-YOU: Thanks for the explanations and implementation. Much better understanding on mutex now.

runtime/list_test.go
trotterdylan: Whoops, yeah you're right. Similar but easier is: `append(Args{l.ToObject()}, args...)`

S-YOU: Thanks

S-YOU:
Updated with mutex changes, and test procedure.

runtime/list_test.go
```diff
@@ -353,6 +353,30 @@ func TestListIteratorIter(t *testing.T) {
...
+	pop := mustNotRaise(GetAttr(NewRootFrame(), ListType.ToObject(), NewStr("pop"), nil))
+	fun := wrapFuncForTest(func(f *Frame, l *List, args ...*Object) (*Tuple, *BaseException) {
+		result, raised := pop.Call(f, append(Args{l.ToObject()}, args...), nil)
+		if raised != nil {
+			return nil, raised
+		}
+		return newTestTuple(result, l), nil
+	})
+	cases := []invokeTestCase{
+		{args: wrapArgs(newTestList(-1, 0, 1)), want: newTestTuple(NewInt(1).ToObject(), newTestList(-1, 0).ToObject()).ToObject()},
```
trotterdylan: You can pass primitive Go types into newTestTuple (and newTestList, newTestDict, wrapArgs, etc.) Also, Grumpy types like *List, *Tuple, etc. don't need to be converted to object to be passed in. This can be:

`newTestTuple(1, newTestList(-1, 0)).ToObject()`

Can make similar simplifications below as well.

runtime/list_test.go
```diff
@@ -353,6 +353,30 @@ func TestListIteratorIter(t *testing.T) {
...
+		{args: wrapArgs(newTestList(-1, 0, 1), 0), want: newTestTuple(NewInt(-1).ToObject(), newTestList(0, 1).ToObject()).ToObject()},
+		{args: wrapArgs(newTestList(-1, 0, 1), NewLong(big.NewInt(1))), want: newTestTuple(NewInt(0).ToObject(), newTestList(-1, 1).ToObject()).ToObject()},
+		{args: wrapArgs(newTestList(-1, 0, 1), None), wantExc: mustCreateException(TypeErrorType, "int() argument must be a string or a number, not 'NoneType'")},
+		{args: wrapArgs(newTestList(-1, 0, 1), None), wantExc: mustCreateException(TypeErrorType, "int() argument must be a string or a number, not 'NoneType'")},
+		{args: wrapArgs(newTestList(-1, 0, 1), 3), wantExc: mustCreateException(IndexErrorType, "list index out of range")},
```
trotterdylan: Worth testing that popping from an empty list produces IndexError and that popping from a single element list returns the empty list.

runtime/list_test.go
S-YOU: Oh, I see. Thank you.

runtime/list_test.go
S-YOU: Sure

S-YOU:
Updated with those changes.

PR #118, opened 2017-01-15
Add _collections to third_party/pypy (deque)

S-YOU:
Added _collections to third_party/pypy, which includes deque and defaultdict. 
deque should works but defaultdict need `__missing__` slot.

```
cat <<EOF > hello.py | make run
import collections

d = collections.deque('ghi')
print d
for elem in d:
  print elem.upper()
d.append('j')
d.appendleft('f')

print d.pop()
print d.popleft()
print list(d)
EOF
```
```
deque(['g', 'h', 'i'])
G
H
I
j
f
['g', 'h', 'i']
```

### DIFF
--- a/lib/_collections.py
+++ b/lib/_collections.py
@@ -1,4 +1,0 @@
-class deque(object):
-
-  def __init__(self, *_):
-    raise NotImplementedError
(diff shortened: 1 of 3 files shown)


S-YOU:
My autopep8 changed the indentation to 2 I think. should I changed back?

S-YOU:
Reverted collections.py as original.

S-YOU:
Reverted _collections.py from pypy with minimal changes too.

PR #124, opened 2017-01-16
Implement str.zfill

S-YOU:


runtime/str.go
```diff
@@ -796,6 +797,28 @@ func strUpper(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return NewStr(strings.ToUpper(s)).ToObject(), nil
 }
 
+func strZFill(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkMethodArgs(f, "zfill", args, StrType, IntType); raised != nil {
```
trotterdylan: Looks like this can accept objects that implement __int__:

```
>>> class A(object):
...  def __int__(self):
...   return 3
...
>>> '3'.zfill(A())
'003'
```

Probably best to check that args[1] is ObjectType and then do `IntType.Call(f, Args{args[1]}, nil)` on it.

S-YOU:
Updated to use IntType.Call, and added more test cases.

trotterdylan:
Good stuff. Thanks for doing this! Merging shortly.

PR #125, opened 2017-01-16
Implement str.find

S-YOU:
I am using new function adjustBounds, without changing existing adjustIndex.
Please let me know your opinion.


runtime/str.go
```diff
@@ -717,6 +757,23 @@ func adjustIndex(i, l int) int {
 	return i
 }
 
+func adjustBounds(f *Frame, o *Object, l int, upperBound bool) (int, *BaseException) {
```
trotterdylan: I'm worried about adding this function because it's not obvious when this should be used vs. adjustIndex. Generally I think the Index/IndexInt/adjustIndex functions are confusing and need to be cleaned up. I would prefer to do this logic inline in strFind rather than adding another function to the mix. You could reuse adjustIndex by doing it as follows:

```
if argc >= 3 {
  start, raised = IndexInt(f, args[2])
  if raised != nil {
    return nil, raised
  }
}
if argc == 4 {
	end, raised = IndexInt(f, args[3])
	if raised != nil {
		return nil, raised
	}
}
if start > end {
	return NewInt(-1).ToObject(), nil
}
start, end = adjustIndex(start, l), adjustIndex(end, l)
```

I may be missing some edge cases here, but the nice thing is that we don't need to add any new helper functions and the semantics are obvious from reading the code.

runtime/str.go
```diff
@@ -205,6 +205,45 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+		expectedTypes = expectedTypes[:argc]
+	}
+	if raised := checkMethodArgs(f, "find/index", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	s := toStrUnsafe(args[0]).Value()
+	l := len(s)
+	start, end := 0, l
+	if argc >= 3 {
+		startParsed, raised := adjustBounds(f, args[2], l, false)
```
trotterdylan: Maybe declare raised above and then just assign directly to start to avoid the extra startParsed temporary:

```
var raised *BaseException
if argc >= 3 {
  start, raised = ...
}
```

runtime/str.go
```diff
@@ -205,6 +205,45 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+		if raised != nil {
+			return nil, raised
+		}
+		end = endParsed
+	}
+	if start > end {
+		return NewInt(-1).ToObject(), nil
+	}
+	sep := toStrUnsafe(args[1]).Value()
+	s = s[start:end]
```
trotterdylan: It's better to avoid reassigning to local variables when possible since it's less obvious to readers what value it has at what point in the function. You can just pass s[start:end] directly to strings.Index() below.

runtime/str.go
S-YOU: Thanks for the implmentation. I added addictional check start > l for the case like `''.find('', 1, 1) `

S-YOU:
Updated based on comments with one additional check.
Also added func doc on strFind, which taken from python [doc](https://docs.python.org/2/library/string.html#string.find).

runtime/str.go
```diff
@@ -205,6 +205,46 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return strCompare(v, w, False, True, False), nil
 }
 
+// strFind returns the lowest index in s where the substring sub is found such
+// that sub is wholly contained in s[start:end]. Return -1 on failure
```
trotterdylan: Period at the end of the second sentence.

runtime/str.go
```diff
@@ -205,6 +205,46 @@ func strEq(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+			return nil, raised
+		}
+	}
+	if argc == 4 {
+		end, raised = IndexInt(f, args[3])
+		if raised != nil {
+			return nil, raised
+		}
+	}
+	if start > end || start > l {
```
trotterdylan: What if end == -1? I think you want to do this check after the adjustIndex calls.

runtime/str.go
S-YOU: Oops. Thanks

S-YOU:
Moved `start > end` after adjustIndex calls, 
`start > l` needs to be there for the case `''.find('', 1, 1) == -1`

runtime/str.go
S-YOU: Missed that test cased, added. Thank you.

PR #126, opened 2017-01-16
Add functools to third_party/pypy

S-YOU:


trotterdylan:
Oh, one thing: I wrote up some basic guidelines for standard library integration: https://github.com/google/grumpy/wiki/Standard-libraries:-guidelines-for-integration

Let me know what you think.

With respect to this change, could we have pulled functools.py from CPython?

S-YOU:
Yeah, make sense. I should move it. Will do in another/other PRs.

cclauss:
Why target 2.7.12 when 2.7.13 is current and has a lot of nice fixes?  https://hg.python.org/cpython/raw-file/v2.7.13/Misc/NEWS

trotterdylan:
Only because 2.7.13 wasn't around when we started bringing in stdlibs. It will be easier to merge in 2.7.12 -> 2.7.13 changes all at once than to have a mix of files.

PR #132, opened 2017-01-16
Avoid panic on types without a dict.

nairb774:


trotterdylan:
I'd rather ensure all types have a dict. Do you know where this is being encountered?

nairb774:
[ObjectType](https://github.com/google/grumpy/blob/de506699345ce1de83381ed50cf39fa59976aa26/runtime/object.go#L31), all [basis types](https://github.com/google/grumpy/blob/de506699345ce1de83381ed50cf39fa59976aa26/runtime/type.go#L131), all [simple types](https://github.com/google/grumpy/blob/de506699345ce1de83381ed50cf39fa59976aa26/runtime/type.go#L140), as well as [TypeType](https://github.com/google/grumpy/blob/de506699345ce1de83381ed50cf39fa59976aa26/runtime/type.go#L330). I'm not sure you want to be able to set extra attributes on the built in types - or maybe that is a thing in py3?

nairb774:
Never mind - ug, you would think I would be more collected. I had found this when poking around in the native code. There was a [test](https://github.com/google/grumpy/blob/de506699345ce1de83381ed50cf39fa59976aa26/runtime/native_test.go#L27) that calls `newNativeType` without adding a dict. Test problem...and a me problem it seems.

nairb774:
Dropping.

trotterdylan:
That test should really be setting a dict on the type. I debated whether it was worth insisting during code review but decided it wasn't worth the trouble. Given it's caused confusion I now see that dict should be set for it.

PR #133, opened 2017-01-17
Add list extend

S-YOU:


runtime/list_test.go
```diff
@@ -220,6 +220,40 @@ func TestListAppend(t *testing.T) {
 	}
 }
 
+func TestListExtend(t *testing.T) {
+	fun := newBuiltinFunction("TestListExtend", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		app, raised := GetAttr(f, ListType.ToObject(), NewStr("extend"), nil)
```
trotterdylan: Might be easier to do this outside of fun:

```
extend := mustNotRaise(GetAttr(NewRootFrame(), ListType.ToObject(), NewStr("extend"), nil)
fun := wrapFuncForTest(func(f *Frame, args ...*Object) (*Object, *BaseException) {
  ...
})

File: runtime/list_test.go [hunk 2]
```diff
@@ -220,6 +220,40 @@ func TestListAppend(t *testing.T) {
...
+		if raised != nil {
+			return nil, raised
+		}
+		if _, raised := app.Call(f, args, nil); raised != nil {
+			return nil, raised
+		}
+		return args[0], nil
+	}).ToObject()
+	cases := []invokeTestCase{
+		{args: wrapArgs(newTestList().ToObject(), newTestTuple().ToObject()), want: newTestList().ToObject()},
```
trotterdylan: No need to convert ToObject() inside of wrapArgs, newTestTuple, newTestList, etc. Those methods take interface{} and do any conversions necessary.

runtime/list.go
```diff
@@ -163,6 +163,14 @@ func listAppend(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return None, nil
 }
 
+func listExtend(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
```
trotterdylan: kwargs is unused so call it `_`

runtime/list.go
```diff
@@ -163,6 +163,14 @@ func listAppend(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	argc := len(args)
+	if argc != 2 {
+		return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("extend() takes exactly one argument (%d given)", argc-1))
```
trotterdylan: If listExtend gets called with empty args (e.g. `list.extend()`), won't this message say "extend() takes exactly one argument (-1 given)"? CPython has a `method` type (distinct from `instancemethod`) that gives this error in that case:

`TypeError: descriptor 'extend' of 'list' object needs an argument`

I don't think we need to worry about implementing `method` right now. Maybe we should settle for having the error message be:

`fmt.Sprintf("extend() takes exactly two arguments (%d given)", argc)`

It's a little weird, but I think less confusing than saying "(-1 given)" in some cases.

runtime/list_test.go
S-YOU: I will keep in mind, thank you very much.

runtime/list.go
S-YOU: list.extend() is okay, I did put in test cases, https://github.com/S-YOU/grumpy/blob/b9cea1251a50e636d90d0f546bb62e71eb6c4494/runtime/list_test.go#L247
which returns `extend() takes exactly one argument (0 given)`
because first argument is list itself as far as I noticed.
Is there is case that unbound method can be called?

I am okay to put argc, but error message is one number off by CPython error message, ok to proceed that?

runtime/list.go
S-YOU: Missed again, thank you.

runtime/list.go
trotterdylan: Yeah unbound methods can be called: list.extend() will do it.

S-YOU: I see, will fix that.

S-YOU:
Missed one changes, will do next.

S-YOU:
Updated those. Thank you.

PR #135, opened 2017-01-17
Implement dict.clear()

S-YOU:
Not sure this is the correct way to clear dictionary, I am just creating newDictTable(0)

### DIFF
--- a/runtime/dict.go
+++ b/runtime/dict.go
@@ -515,6 +515,7 @@ func dictClear(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	d := toDictUnsafe(args[0])
 	d.mutex.Lock(f)
 	d.table = newDictTable(0)
+	d.incVersion()
 	d.mutex.Unlock(f)
 	return None, nil
 }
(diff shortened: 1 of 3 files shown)


runtime/dict.go
```diff
@@ -508,6 +508,17 @@ func dictsAreEqual(f *Frame, d1, d2 *Dict) (bool, *BaseException) {
 	return result, nil
 }
 
+func dictClear(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkMethodArgs(f, "clear", args, DictType); raised != nil {
+		return nil, raised
+	}
+	d := toDictUnsafe(args[0])
+	d.mutex.Lock(f)
+	d.table = newDictTable(0)
```
trotterdylan: Also do incVersion()

S-YOU:
Added incVersion(), thank you.

runtime/dict.go
S-YOU: Done that, I hope position I put is ok.

PR #142, opened 2017-01-17
Add more test cases from stdlib

S-YOU:
list, dict, set, str, unicode, int, slice

```
./run third_party/stdlib/test/test_list.py
test_basic (__main__.ListTest) ... ok
test_identity (__main__.ListTest) ... ok
test_len (__main__.ListTest) ... ok
test_overflow (__main__.ListTest) ... ok
test_truth (__main__.ListTest) ... ok

Ran 5 tests in 0.001133s

OK
```

```
./run third_party/stdlib/test/test_dict.py                                                                                                                                     [master]$
test_bad_key (__main__.DictTest) ... ERROR
test_bool (__main__.DictTest) ... ok
test_clear (__main__.DictTest) ... ERROR
test_constructor (__main__.DictTest) ... ok
test_container_iterator (__main__.DictTest) ... ERROR
test_contains (__main__.DictTest) ... ok
test_copy (__main__.DictTest) ... ERROR
test_empty_presized_dict_in_freelist (__main__.DictTest) ... ok
test_fromkeys (__main__.DictTest) ... ERROR
test_get (__main__.DictTest) ... ok
test_getitem (__main__.DictTest) ... ok
test_has_key (__main__.DictTest) ... ERROR
test_items (__main__.DictTest) ... ok
test_keys (__main__.DictTest) ... ERROR
test_le (__main__.DictTest) ... FAIL
test_len (__main__.DictTest) ... ok
test_literal_constructor (__main__.DictTest) ... ERROR
test_missing (__main__.DictTest) ... ERROR
test_mutatingiteration (__main__.DictTest) ... ok
test_pop (__main__.DictTest) ... ERROR
test_popitem (__main__.DictTest) ... ERROR
test_repr (__main__.DictTest) ... ok
test_resize1 (__main__.DictTest) ... ok
test_resize2 (__main__.DictTest) ... ok
test_setdefault (__main__.DictTest) ... ERROR
test_setdefault_atomic (__main__.DictTest) ... ERROR
test_tuple_keyerror (__main__.DictTest) ... ERROR
test_update (__main__.DictTest) ... ERROR
test_values (__main__.DictTest) ... ERROR
...
Ran 29 tests in 0.010734s

FAILED (failures=1, errors=16)
TestFailed: multiple errors occurred
```

...

### DIFF
--- a/third_party/stdlib/test/list_tests.py
+++ b/third_party/stdlib/test/list_tests.py
@@ -1,0 +1,545 @@
+"""
+Tests common to list and UserList.UserList
+"""
+
+import sys
+import os
+
+from test import test_support, seq_tests
+
+class CommonTest(seq_tests.CommonTest):
+
+    def test_init(self):
+        # Iterable arg is optional
+        self.assertEqual(self.type2test([]), self.type2test())
+
+        # Init clears previous values
+        a = self.type2test([1, 2, 3])
+        a.__init__()
+        self.assertEqual(a, self.type2test([]))
+
+        # Init overwrites previous values
+        a = self.type2test([1, 2, 3])
+        a.__init__([4, 5, 6])
+        self.assertEqual(a, self.type2test([4, 5, 6]))
+
+        # Mutables always return a new object
+        b = self.type2test(a)
+        self.assertNotEqual(id(a), id(b))
+        self.assertEqual(a, b)
+
+    def test_repr(self):
+        l0 = []
+        l2 = [0, 1, 2]
+        a0 = self.type2test(l0)
+        a2 = self.type2test(l2)
+
+        self.assertEqual(str(a0), str(l0))
+        self.assertEqual(repr(a0), repr(l0))
+        self.assertEqual(repr(a2), repr(l2))
+        self.assertEqual(str(a2), "[0, 1, 2]")
+        self.assertEqual(repr(a2), "[0, 1, 2]")
+
+        a2.append(a2)
+        a2.append(3)
+        self.assertEqual(str(a2), "[0, 1, 2, [...], 3]")
+        self.assertEqual(repr(a2), "[0, 1, 2, [...], 3]")
+
+        l0 = []
+        for i in xrange(sys.getrecursionlimit() + 100):
+            l0 = [l0]
+        self.assertRaises(RuntimeError, repr, l0)
+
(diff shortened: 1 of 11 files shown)


S-YOU:
I only commented out those does not compile, but not error cases. First list list is only doing part of the list, thats why it passed I think.

S-YOU:
By the way there is one panic on unicode test, when I last ran.

S-YOU:
In your opinion, only `third_party/stdlib/test/test_list.py` should be in now?
Already merged third_party/stdlib/test/test_tuple.py didn't passed all tests so, may be I will do another PR, with one addition and one deletion, let me know if ok to proceed.

trotterdylan:
I'm fine with commenting out code that doesn't yet compile. I'm also fine with leaving in failing tests. But I think it makes sense to mark the failing tests with expectedFailure so that the whole suite doesn't fail. It's not clear to me whether tests that error will be squelched by expectedFailure. If they still cause exceptions to be raised when they are marked as expectedFailure then we could mark them with unittest.skip so that they're present in the suite but skipped until we can fix them.

The end result should be that test suites that are checked in pass and we can uncomment or remove the skip/expectedFailure tags as we get more tests working.

Does that make sense?

S-YOU:
`expectedFailure` you mean is that grumpy does not supported yet but CPython may return sucess? 

Not like following
```
class ExpectedFailureTestCase(unittest.TestCase):
    @unittest.expectedFailure
    def test_fail(self):
        self.assertEqual(1, 0, "broken")
```

Sorry about asking, I am actually new to those things.

trotterdylan:
No worries, I've never used expectedFailure myself, so this is new to me too :)

What you've described is the same as what I was thinking. Tests that fail for Grumpy but pass for CPython, we mark with expectedFailure like:

```
  def test_something(self):
    ...
  test_something = unittest.expectedFailure(test_something)
```

(Note since decorators don't yet work, we have to call expectedFailure directly.)

Tests that are marked like this will still be run but will not cause the whole suite to fail so our tests can "pass". That way we know when we break something that used to work, or we fix something that used to be broken.

S-YOU:
Got it, Thank you very much, for that case I think I better submit each type/module per PR.

S-YOU:
Closing for now..

PR #145, opened 2017-01-18
Chore third party libs (use stdlib whenever possible)

S-YOU:
- Moved re modules to stdlib except _sre.py
- Moved md5.py to stdlib
- Moved functools.py to stdlib

Also chore a bit to make minimal changes from CPython.


S-YOU:
Rebased.

PR #151, opened 2017-01-18
Make acceptance tests pass under CPython.

trotterdylan:
All tests in testing/*_test.py should pass in CPython and Grumpy (except native_test.py since that tests Grumpy native extensions). This change makes it so that these tests run under CPython as part of `make test`.

Mostly, things were already passing, but a few code paths result in different exception types or messages. Differing messages is acceptable but the types should be the same, so those parts need to be fixed. I added TODOs for those code paths which all had to do with how str.find() start/end args are handled.

### DIFF
--- a/testing/str_test.py
+++ b/testing/str_test.py
@@ -61,11 +61,8 @@ class Foo(object):
     return 3
 assert 'abcd'.find('a', Foo()) == -1
 
-try:
-  'ab'.find('xxx', sys.maxsize + 1, 0)
-  raise AssertionError
-except IndexError:
-  pass
+# TODO: This raises IndexError under Grumpy but returns -1 for CPython.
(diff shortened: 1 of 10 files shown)


S-YOU:
Thanks for changes, I realized that after I saw CPython tests file that they are only checking Type of Error, not messages.

testing/str_test.py
```diff
@@ -85,17 +82,19 @@ def __index__(self):
...
-  raise AssertionError
-except IndexError:
-  pass
+# TODO: Both of these test cases raise TypeError under CPython but raise
+# IndexError under Grumpy.
+#try:
+#  'foobar'.find("bar", "baz")
+#  raise AssertionError
+#except TypeError:
+#  pass
```
S-YOU: Looks like my nil check on IndexInt is incomplete. Will fix.

PR #152, opened 2017-01-18
Simplify with or,  msg lookup with dict.get()

cclauss:
Three minor optimizations

### DIFF
--- a/compiler/block.py
+++ b/compiler/block.py
@@ -33,11 +33,9 @@ class Package(object):
 
   def __init__(self, name, alias=None):
     self.name = name
-    if not alias:
       # Use Γ as a separator since it provides readability with a low
       # probability of name collisions.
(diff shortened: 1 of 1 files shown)


compiler/block.py
```diff
@@ -413,18 +411,15 @@ def _assign_target(self, target):
   def _register_global(self, node, name):
     var = self.vars.get(name)
     if var:
-      if var.type == Var.TYPE_PARAM:
-        msg = "name '{}' is parameter and global"
-        raise util.ParseError(node, msg.format(name))
-      if var.type == Var.TYPE_LOCAL:
-        msg = "name '{}' is used prior to global declaration"
+      msg = {Var.TYPE_LOCAL: "name '{}' is used prior to global declaration",
```
trotterdylan: While this is fewer lines, I find that burying the logic in the get() call actually makes the code's intent less obvious.

PR #154, opened 2017-01-18
Add install section

xyproto:
Added an `install` section to the Makefile.

It's not perfect, but it appears to work for [this unofficial Arch Linux package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=grumpy).

This change allows grumpy to be installed on the system and makes it easier to package for Linux maintainers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,3 +244,19 @@ $(patsubst %,build/%.go,$(BENCHMARKS)): build/%.go: %.py $(COMPILER)
 $(BENCHMARK_BINS): build/benchmarks/%_benchmark: build/benchmarks/%.go $(RUNTIME) $(STDLIB)
 	@mkdir -p $(@D)
 	@go build -o $@ $<
+
+# ------------------------------------------------------------------------------
+# Installation
+# ------------------------------------------------------------------------------
+
+install:
+	install -d $(DESTDIR)/usr/lib/go
+	install -d $(DESTDIR)/usr/bin
+	install -Dm755 tools/benchcmp $(DESTDIR)/usr/bin/benchcmp
+	install -Dm755 tools/coverparse $(DESTDIR)/usr/bin/coverparse
+	install -Dm755 tools/diffrange $(DESTDIR)/usr/bin/diffrange
+	install -Dm755 tools/grumpc $(DESTDIR)/usr/bin/grumpc
+	install -Dm755 tools/grumprun $(DESTDIR)/usr/bin/grumprun
+	cp -rv --no-preserve=ownership build/bin build/lib $(DESTDIR)/usr/
+	cp -rv --no-preserve=ownership build/pkg build/src $(DESTDIR)/usr/lib/go/
+
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ test: $(ACCEPT_PASS_FILES) $(ACCEPT_PY_PASS_FILES) $(COMPILER_PASS_FILES) $(COMP
 
 precommit: cover gofmt lint test
 
-.PHONY: all benchmarks clean cover gofmt golint lint precommit pylint run test install
+.PHONY: all benchmarks clean cover gofmt golint install lint precommit pylint run test
 
 # ------------------------------------------------------------------------------
 # grumpc compiler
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -227,3 +227,19 @@ $(patsubst %,build/%.go,$(BENCHMARKS)): build/%.go: %.py $(COMPILER)
...
 	@go build -o $@ $<
+
+# ------------------------------------------------------------------------------
+# Installation
+# ------------------------------------------------------------------------------
+
+install:
+	install -d $(DESTDIR)/usr/lib/go
+	install -d $(DESTDIR)/usr/bin
+	install -Dm755 tools/benchcmp $(DESTDIR)/usr/bin/benchcmp
```
trotterdylan: I wouldn't worry about including all the binaries, just grumpc and grumprun.

Makefile
```diff
@@ -227,3 +227,19 @@ $(patsubst %,build/%.go,$(BENCHMARKS)): build/%.go: %.py $(COMPILER)
 $(BENCHMARK_BINS): build/benchmarks/%_benchmark: build/benchmarks/%.go $(RUNTIME) $(STDLIB)
 	@mkdir -p $(@D)
 	@go build -o $@ $<
+
+# ------------------------------------------------------------------------------
+# Installation
+# ------------------------------------------------------------------------------
+
+install:
```
trotterdylan: This rule should depend on $(RUNNER_BIN) $(COMPILER) $(RUNTIME) $(STDLIB) so that all the necessary files are present in the build subdir.

trotterdylan: Add `install` to the `.PHONY` prerequisites.

Makefile
```diff
@@ -227,3 +227,19 @@ $(patsubst %,build/%.go,$(BENCHMARKS)): build/%.go: %.py $(COMPILER)
...
+	install -d $(DESTDIR)/usr/lib/go
+	install -d $(DESTDIR)/usr/bin
+	install -Dm755 tools/benchcmp $(DESTDIR)/usr/bin/benchcmp
+	install -Dm755 tools/coverparse $(DESTDIR)/usr/bin/coverparse
+	install -Dm755 tools/diffrange $(DESTDIR)/usr/bin/diffrange
+	install -Dm755 tools/grumpc $(DESTDIR)/usr/bin/grumpc
```
trotterdylan: You should copy the tools from build/bin since the shebang line for those scripts gets hardcoded to whatever $(PYTHON) is during the build step.

xyproto:
Thanks for the comments, will update the pull request.

Makefile
```diff
@@ -227,3 +227,15 @@ $(patsubst %,build/%.go,$(BENCHMARKS)): build/%.go: %.py $(COMPILER)
...
+
+# ------------------------------------------------------------------------------
+# Installation
+# ------------------------------------------------------------------------------
+
+install: $(RUNNER_BIN) $(COMPILER) $(RUNTIME) $(STDLIB)
+	install -Dm755 build/bin/grumpc $(DESTDIR)/usr/bin/grumpc
+	install -Dm755 build/bin/grumprun $(DESTDIR)/usr/bin/grumprun
+	install -d $(DESTDIR)/usr/lib/go
+	cp -rv --no-preserve=ownership build/lib $(DESTDIR)/usr/
```
trotterdylan: Actually I don't think we want to install everything in lib since it includes some stuff like pylint that isn't actually required for grumpy. Basically all we want is what's in site-packages/grumpy. And that should end up in the pythonpath for the selected PYTHON bin... Can we get the prefix for that Python and then put it in lib in there?

xyproto: Yes, it's possible to find the correct site-path for the specified Python. It requires some fiddling about, but it is doable. Will fix.

xyproto:
Yes.

xyproto:
Grumpy is now an [official Arch Linux package](https://www.archlinux.org/packages/community/x86_64/grumpy/). Cheers!

trotterdylan:
Very cool. Thanks for helping with that!

cclauss:
See https://github.com/google/grumpy/issues/210

PR #155, opened 2017-01-18
Fix IndexError to TypeError

S-YOU:
- Supposed to fix IndexError to TypeError except int, long or with `__index__` method.
- Added passing None to start or end, which supposed to take as default values.
- Note that sys.maxsize + 1 issue is not yet fixed.

### DIFF
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -501,7 +501,7 @@ func IndexInt(f *Frame, o *Object) (i int, raised *BaseException) {
 	if o.isInstance(LongType) {
 		l := toLongUnsafe(o).Value()
 		// Anything bigger than maxIntBig will treat as maxIntBig.
-		if maxIntBig.Cmp(l) < 0 {
+		if !numInIntRange(l) {
 			l = maxIntBig
 		}
 		return int(l.Int64()), nil
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -487,22 +487,26 @@ func Index(f *Frame, o *Object) (*Object, *BaseException) {
 // IndexInt returns the value of o converted to a Go int according to o's
 // __index__ slot.
 // It raises a TypeError if o doesn't have an __index__ method.
-// If the returned value doesn't fit into an int, it raises an OverflowError.
-func IndexInt(f *Frame, o *Object) (int, *BaseException) {
-	i, raised := Index(f, o)
-	if raised != nil {
-		return 0, raised
+func IndexInt(f *Frame, o *Object) (i int, raised *BaseException) {
+	if index := o.typ.slots.Index; index != nil {
+		// Unwrap __index__ slot and fall through.
+		o, raised = index.Fn(f, o)
+		if raised != nil {
+			return 0, raised
+		}
 	}
-	if i == nil {
-		return 0, f.RaiseType(TypeErrorType, "slice indices must be integers or None or have an __index__ method")
+	if o.isInstance(IntType) {
+		return toIntUnsafe(o).Value(), nil
 	}
-	if i.isInstance(IntType) {
-		return toIntUnsafe(i).Value(), nil
-	}
-	if l := toLongUnsafe(i).Value(); numInIntRange(l) {
+	if o.isInstance(LongType) {
+		l := toLongUnsafe(o).Value()
+		// Anything bigger than maxIntBig will treat as maxIntBig.
+		if maxIntBig.Cmp(l) < 0 {
+			l = maxIntBig
+		}
 		return int(l.Int64()), nil
 	}
-	return 0, f.RaiseType(IndexErrorType, fmt.Sprintf("cannot fit '%s' into an index-sized integer", o.typ))
+	return 0, f.RaiseType(TypeErrorType, errBadSliceIndex)
 }
 
 // Invoke calls the given callable with the positional arguments given by args
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -493,13 +493,14 @@ func IndexInt(f *Frame, o *Object) (int, *BaseException) {
 	if raised != nil {
 		return 0, raised
 	}
-	if i != nil {
-		if i.isInstance(IntType) {
-			return toIntUnsafe(i).Value(), nil
-		}
-		if l := toLongUnsafe(i).Value(); numInIntRange(l) {
-			return int(l.Int64()), nil
-		}
+	if i == nil || (!i.isInstance(IntType) && !i.isInstance(LongType)) {
+		return 0, f.RaiseType(TypeErrorType, "slice indices must be integers or None or have an __index__ method")
+	}
+	if i.isInstance(IntType) {
+		return toIntUnsafe(i).Value(), nil
+	}
+	if l := toLongUnsafe(i).Value(); numInIntRange(l) {
+		return int(l.Int64()), nil
 	}
 	return 0, f.RaiseType(IndexErrorType, fmt.Sprintf("cannot fit '%s' into an index-sized integer", o.typ))
 }
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -493,7 +493,7 @@ func IndexInt(f *Frame, o *Object) (int, *BaseException) {
 	if raised != nil {
 		return 0, raised
 	}
-	if i == nil || (!i.isInstance(IntType) && !i.isInstance(LongType)) {
+	if i == nil {
 		return 0, f.RaiseType(TypeErrorType, "slice indices must be integers or None or have an __index__ method")
 	}
 	if i.isInstance(IntType) {
--- a/runtime/slice.go
+++ b/runtime/slice.go
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
 	SliceType.slots.Repr = &unaryOpSlot{sliceRepr}
 }
 
+func sliceIndex(f *Frame, o *Object) (i int64, raised *BaseException) {
+	if index := o.typ.slots.Index; index != nil {
+		o, raised = index.Fn(f, o)
+		if raised != nil {
+			return int64(0), raised
+		}
+	}
+	if o.isInstance(IntType) {
+		return int64(toIntUnsafe(o).Value()), nil
+	}
+	if o.isInstance(LongType) {
+		return toLongUnsafe(o).Value().Int64(), nil
+	}
+	return 0, f.RaiseType(TypeErrorType, errBadSliceIndex)
+}
+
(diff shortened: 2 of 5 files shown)


runtime/core.go
```diff
@@ -493,13 +493,14 @@ func IndexInt(f *Frame, o *Object) (int, *BaseException) {
...
 		return 0, raised
 	}
-	if i != nil {
-		if i.isInstance(IntType) {
-			return toIntUnsafe(i).Value(), nil
-		}
-		if l := toLongUnsafe(i).Value(); numInIntRange(l) {
-			return int(l.Int64()), nil
-		}
+	if i == nil || (!i.isInstance(IntType) && !i.isInstance(LongType)) {
```
trotterdylan: I think things can be simplified somewhat because int and long both have the `__index__` slot. I believe Index() checks whether o is int or long as an optimization, but perhaps it's better if we just remove the check on 469.

That will guarantee that either we already raised or `i == nil || i.isInstance(IntType) || i.isInstance(LongType))` so we can remove the second part of the condition here.

runtime/str.go
```diff
@@ -221,13 +221,13 @@ func strFind(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	s := toStrUnsafe(args[0]).Value()
 	l := len(s)
 	start, end := 0, l
-	if argc >= 3 {
+	if argc >= 3 && args[2] != None {
```
trotterdylan: It occurred to me when I saw that CPython gives an error message about a slice index that they must be using some helper function related to slices. We have a similar helper, sliceClampIndex(), that handles None, checks for `__index__`, etc. I think we could just use that, something like:

```
if argc >= 3 {
  start, raised = sliceClampIndex(f, args[2], 0, l)
  if raised != nil {
    return nil, raised
  }
}
```

S-YOU: Similar to adjustIndex, this block
```
        if i > seqLen {
		i = seqLen
	}
```
Fails for `"".find("", 1)` case, I think adjusting index for start and end can't be together.

runtime/core.go
S-YOU: Thanks, I noticed that things get complicated there, but couldn't figured out.

S-YOU:
@trotterdylan, I have read CPython source code of PySlice_GetIndicesEx, and _PyEval_SliceIndex,
and come up with new function sliceGetIndices, what do you think?
It pass all existing tests, including sys.maxsize + 1.

One coverage missing because I don't know how to write following in Go.
```
class Foo(object):
  def __index__(self):
    return 3
```

runtime/slice.go
```diff
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
 	SliceType.slots.Repr = &unaryOpSlot{sliceRepr}
 }
 
+func sliceIndex(f *Frame, o *Object) (i int64, raised *BaseException) {
```
trotterdylan: Prefer using int to int64 when possible. For the long case, you can compare its value to maxIntBig and minIntBig to make sure it's in range.

runtime/slice.go
```diff
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
...
+	if index := o.typ.slots.Index; index != nil {
+		o, raised = index.Fn(f, o)
+		if raised != nil {
+			return int64(0), raised
```
trotterdylan: Is the int64() specification necessary here? I thought the constant 0 would be typeless.

runtime/slice.go
```diff
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
...
+			stop += length
+		}
+		if stop < 0 {
+			stop = 0
+		}
+		if stop >= length {
+			stop = length
+		}
+	}
+	return
```
trotterdylan: Remove return

runtime/slice.go
```diff
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
...
 }
 
+func sliceIndex(f *Frame, o *Object) (i int64, raised *BaseException) {
+	if index := o.typ.slots.Index; index != nil {
+		o, raised = index.Fn(f, o)
+		if raised != nil {
+			return int64(0), raised
+		}
+	}
+	if o.isInstance(IntType) {
```
trotterdylan: I think this condition and the LongType condition below will never be true, since both IntType and LongType have the `__index__` slot.

runtime/slice.go
trotterdylan: The name sliceIndex could be confused with the slice type's `__index__` slot so I think it's worth coming up with a different name.

runtime/slice.go
```diff
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
...
+		return int64(toIntUnsafe(o).Value()), nil
+	}
+	if o.isInstance(LongType) {
+		return toLongUnsafe(o).Value().Int64(), nil
+	}
+	return 0, f.RaiseType(TypeErrorType, errBadSliceIndex)
```
trotterdylan: This function is very similar to Index() isn't it? What distinguishes when Index() should be used vs. sliceIndex()?

runtime/slice.go
```diff
@@ -156,6 +156,58 @@ func initSliceType(map[string]*Object) {
...
+}
+
+func sliceGetIndices(f *Frame, oStart, oStop *Object, length int64) (start, stop int64, raised *BaseException) {
```
trotterdylan: Will this method be used by slices? Should this replace some of the logic in calcSlice?

runtime/slice.go
S-YOU: Oops, yes, you're right.

runtime/slice.go
S-YOU: maxIntBig, minIntBig are 2147483647,-2147483648
While sys.maxsize, and sys.maxsize + 1 is 9223372036854775807, 9223372036854775808L
Is there a way to check it without using int64?

runtime/slice.go
S-YOU: You're right, I will reorder the checks.

runtime/slice.go
trotterdylan: That's odd. maxIntBig is [defined as grumpy.MaxInt](https://github.com/google/grumpy/blob/master/runtime/numeric.go#L35). sys.maxsize is [defined as sys.maxint](https://github.com/google/grumpy/blob/master/lib/sys.py#L28), which is `__go__.grumpy.MaxInt`. So maxIntBig should be the same as sys.maxsize.

runtime/slice.go
S-YOU: One difference is this check in Index, apart from Index returning *Object
```
index := o.typ.slots.Index
	if index == nil {
		return nil, nil
	}
```
it return nil, nil, which need extra check outside.

runtime/slice.go
S-YOU: Very similar, but I haven't think about other use cases yet, because new function still in proposal state.

runtime/slice.go
trotterdylan: I see. They are very similar though. It will be difficult for people to understand which one to use. Is there a way we can structure things so that one of these functions can go away?

runtime/slice.go
S-YOU: Go does not compile somehow with that.

trotterdylan: Oh right, because you have named return values. Sorry for the noise.

runtime/slice.go
S-YOU: Ah, sorry, I copy pasted that code in golang.org and tried that, got those results.
https://play.golang.org/p/KLg6Io_tO5
grumpy implementation is okay

runtime/slice.go
S-YOU: I actually didn't know, there is difference with named and without. Thanks for the info.

runtime/slice.go
S-YOU: Yeah, I will try to work to make one general use case for both. Hopefully without hitting corner cases.

runtime/slice.go
S-YOU: Looks like 9223372036854775808L  not in int range, 
sys.maxsize + 1 pretty much corner case with no actual use case.
Do we support it? If so, I need to add more checks with big.Int.

trotterdylan: Since CPython handles it, I think we should handle it as well. One relatively simple way to do it for *Long values would be to do maxIntBig.Cmp(start.Value()) and always return -1 if the pos param is bigger than maxIntBig.

How does CPython implement this check?

S-YOU: Looks like Python only check as Subclass of Int and the actual PyIntObject store the value of Int in long, and cast it with [PyInt_AS_LONG](https://github.com/python/cpython/blob/2.7/Python/ceval.c#L4694).
```
typedef struct {
    PyObject_HEAD
    long ob_ival;
} PyIntObject;
```

S-YOU: return -1 from that function is not possible because it get later got adjusted, is it ok to return maxIntBig instead?

runtime/slice.go
S-YOU: Actually above index.Fn return the object and fall though, so they are needed.
Instead I wrote a comment instead, since it looks like confusion can come from reassigning variable `o`.

runtime/slice.go
S-YOU: I am still finding way to merge with existing `Index`

S-YOU:
Instead of adding new functions, I have modified IndexInt and adjustIndex functions.
adjustIndex logic is basically copy from [CPython](https://github.com/python/cpython/blob/2.7/Objects/stringlib/find.h#L51)

Coverage is still yet to be confirmed.

runtime/slice.go
trotterdylan: Sorry, I meant we need to return -1 from strFind in that case.

runtime/core.go
```diff
@@ -487,21 +487,26 @@ func Index(f *Frame, o *Object) (*Object, *BaseException) {
...
-		if l := toLongUnsafe(i).Value(); numInIntRange(l) {
-			return int(l.Int64()), nil
+	}
+	if o.isInstance(IntType) {
+		return toIntUnsafe(o).Value(), nil
+	}
+	if o.isInstance(LongType) {
+		l := toLongUnsafe(o).Value()
+		// Anything bigger than maxIntBig will treat as maxIntBig.
+		if maxIntBig.Cmp(l) < 0 {
```
trotterdylan: You can use numInIntRange() for this.

S-YOU: Updated that, I thought sys.maxsize+1 is not in the range.

PR #162, opened 2017-01-18
Implement the 'reverse' option for sorted()

MirkoDziadzka:
Add the 'reverse' option to 'sorted()'

I'm not sure if this violates the 'stable sort definition in Python. 
On the other hand, this is probably only relevant when key or cmp gets implemented.


runtime/builtin_types.go
```diff
@@ -580,6 +580,14 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 		return nil, raised
 	}
 	toListUnsafe(result).Sort(f)
+	// Implement reverse. 
+	reverse, raised := IsTrue(f, kwargs.get("reverse", None))
+	if raised != nil {
+		return nil, raised
+	}
+	if reverse {
+		toListUnsafe(result).Reverse()
```
trotterdylan: As you mention, stability of the results is a concern here. If we supported the cmp param, we could negate the result of that to achieve reversal. This would mean that listSorter would need to support cmp. Then sorted(... reverse=True) could wrap that cmp and negate the result (or provide a stock reverse cmp when cmp=None).

So maybe it makes sense to implement cmp first?

trotterdylan (changes requested):
Thanks for working on this! I have a couple high level comments.

runtime/builtin_types.go
```diff
@@ -570,7 +570,7 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return s.ToObject(), nil
 }
 
-func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+func builtinSorted(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
```
MirkoDziadzka: Maybe we should not implement this in sorted(), but in listSort and just change the sorted implementation here to call

```diff
-func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+func builtinSorted(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
        // TODO: Support (cmp=None, key=None, reverse=False)
        if raised := checkFunctionArgs(f, "sorted", args, ObjectType); raised != nil {
                return nil, raised
@@ -579,7 +579,13 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
        if raised != nil {
                return nil, raised
        }
-       toListUnsafe(result).Sort(f)
+       // Creating newArgs by replacing the first element with teh new list.
+       newArgs := args.makeCopy()
+       newArgs[0] = result
+       _, raised = listSort(f, newArgs, kwargs)
+       if raised != nil {
+               return nil, raised
+       }
        return result, nil
 }
```

runtime/builtin_types.go
MirkoDziadzka: Ack. But I would prefer to implement this in listSort and changing sorted() to

```diff
-func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+func builtinSorted(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
        // TODO: Support (cmp=None, key=None, reverse=False)
        if raised := checkFunctionArgs(f, "sorted", args, ObjectType); raised != nil {
                return nil, raised
@@ -579,7 +579,13 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
        if raised != nil {
                return nil, raised
        }
-       toListUnsafe(result).Sort(f)
+       // Creating newArgs by replacing the first element with teh new list.
+       newArgs := args.makeCopy()
+       newArgs[0] = result
+       _, raised = listSort(f, newArgs, kwargs)
+       if raised != nil {
+               return nil, raised
+       }
        return result, nil
 }
```

runtime/builtin_types.go
trotterdylan: SGTM

trotterdylan:
That all sounds good. Let me know when I should review again.

PR #167, opened 2017-01-19
Add ToInt() function for converting int args

trotterdylan:
Also fix some unsafe conversions.


S-YOU:
Will this be replacement for IntType.Call?

trotterdylan:
Yes, I realized that IntType.Call() isn't quite right since you could provide a string value which would be implicitly converted to a number. That's not how CPython does it.

PR #168, opened 2017-01-19
Implement str.replace

S-YOU:
Implementing str.replace.
This will have some updates from #167, so this is WIP for now.


runtime/str.go
```diff
@@ -453,6 +453,37 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
 }
 
+// strReplace returns a copy of the string s with the first n non-overlapping
+// instances of old replaced by new. If old is empty, it matches at the
+// beginning of the string. If n < 0, there is no limit on the number of
+// replacements.
+func strReplace(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	// TODO: Support unicode replace.
+	expectedTypes := []*Type{StrType, StrType, StrType, ObjectType}
+	argc, n := len(args), -1
```
trotterdylan: It's better not to group variables that are unrelated since it's harder to find their definition.

runtime/str.go
```diff
@@ -453,6 +453,37 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	if argc == 4 {
+		o, raised := IntType.Call(f, Args{args[3]}, nil)
+		if raised != nil {
+			return nil, raised
+		}
+		n = toIntUnsafe(o).Value()
+	}
+	s := toStrUnsafe(args[0]).Value()
+	old := toStrUnsafe(args[1]).Value()
+	new := toStrUnsafe(args[2]).Value()
```
trotterdylan: Try to avoid using names that collide with builtin names.

runtime/str.go
```diff
@@ -453,6 +453,37 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	// CPython only, pypy supposed to be same as Go
```
trotterdylan: Can you elaborate on this comment?

runtime/str.go
```diff
@@ -453,6 +453,37 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	if len(s) == 0 && len(old) == 0 && n >= 0 {
+		return NewStr("").ToObject(), nil
+	}
+	return NewStr(strings.Replace(s, old, new, n)).ToObject(), nil
```
trotterdylan: I think strings.Replace() may not be compatible with the Python version because when `old == ""`, strings.Replace: `matches at the beginning of the string and after each UTF-8 sequence` whereas in Python it will match after each *byte*. So if the string contains UTF-8 sequences, the results will differ (this is a good test case to have, as well).

runtime/str.go
S-YOU: I haven't really tested on pypy, but looks like it is different.
http://bugs.python.org/issue28029
assert ''.replace('', 'x') == 'x'
assert ''.replace('', 'x', 1000) == ''

runtime/str.go
S-YOU: Thank you, make sense.

runtime/str.go
S-YOU: Sure, too much copying from CPython.

runtime/str.go
S-YOU: Oh, I haven't think about UTF-8 yet, also unicode.

S-YOU:
- Rebased, fix reviewed issues
- Added utf8 replace test cases, but not sure I am testing real issues.

runtime/str.go
S-YOU: I am not very sure which utf8 sequence is best to test, but added some.

trotterdylan:
Thanks for the changes! The issue I'm pointing out is that in Go strings.Replace() treats the input as a UTF-8 encoded string and if the old string is `""` then it will add a replacement in between each UTF-8 code point. E.g.:

```
package main

import (
	"fmt"
	"strings"
)

func main() {
	fmt.Println([]byte(strings.Replace("\xd0\xb2\xd0\xbe\xd0\xbb", "", "\x00", -1)))
}
```

outputs: [0 208 178 0 208 190 0 208 187 0]

vs. CPython

```
>>> [ord(c) for c in"\xd0\xb2\xd0\xbe\xd0\xbb".replace('', '\0')]
[0, 208, 0, 178, 0, 208, 0, 190, 0, 208, 0, 187, 0]
```

I don't think it's a good idea to use strings.Replace() because of these subtle incompatibilities.

S-YOU:
Hmm, I see. Is that mean, need to implement by looping through bytes?

trotterdylan:
Looks like there's a bytes replace: https://golang.org/pkg/bytes/#Replace

S-YOU:
Thank you but it mention same thing, 
`If old is empty, it matches at the beginning of the slice and after each UTF-8 sequence`

The issue is only on empty old string? then I can just write a loop for that.

trotterdylan:
It's probably worth looking at the go source code to make sure that's the
only peculiarity. But yeah maybe that's the best approach.

S-YOU:
Updated. I checked both strings.Replace and bytes.Replace source. It appears that they doing special checking on len(old) == 0, so I added a loop to copy manually for len(old) case. 
	            if len(old) == 0 {
			if i > 0 {
				_, wid := utf8.DecodeRuneInString(s[start:])
				j += wid
			}
I was thinking to copy codes from go source, but a lot of code to copy on.

runtime/str.go
```diff
@@ -453,6 +453,65 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	if argc == 4 {
+		n, raised = ToIntValue(f, args[3])
+		if raised != nil {
+			return nil, raised
+		}
+	}
+	s := toStrUnsafe(args[0]).Value()
+	old := toStrUnsafe(args[1]).Value()
+	sub := toStrUnsafe(args[2]).Value()
+	// CPython only, pypy supposed to be same as Go.
```
trotterdylan: I'm a little confused about the meaning of this comment. Could you try to make the message a little clearer?

runtime/str.go
```diff
@@ -453,6 +453,65 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	if len(s) == 0 && len(old) == 0 && n >= 0 {
```
trotterdylan: I'm probably overlooking something but don't we always want to return "" when len(s) is zero?

runtime/str.go
```diff
@@ -453,6 +453,65 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	if len(old) == 0 {
+		if n < 0 {
+			n = len(s) + 1
+		}
+		buf := bytes.Buffer{}
+		if n > 0 {
+			buf.WriteString(sub)
+			n--
+		}
+		for i := 0; i < len(s); i++ {
```
trotterdylan: Store len(s) as a temporary numBytes or something so it's not recomputed every iteration.

runtime/str.go
```diff
@@ -453,6 +453,65 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+	old := toStrUnsafe(args[1]).Value()
+	sub := toStrUnsafe(args[2]).Value()
+	// CPython only, pypy supposed to be same as Go.
+	if len(s) == 0 && len(old) == 0 && n >= 0 {
+		return NewStr("").ToObject(), nil
+	}
+	if n == 0 {
+		return NewStr(s).ToObject(), nil
+	}
+	if len(old) == 0 {
```
trotterdylan: Maybe switch the conditional to `len(old) > 0` and do the strings.Replace() call to exit the function early. That would mean this big block of code doesn't have to be indented and it's probably easier to read.

trotterdylan: Worth adding a comment about why we do all this extra work instead of using strings.Replace()

trotterdylan: Prefer `old == ""` to `len(old) == 0`

runtime/str.go
```diff
@@ -453,6 +453,65 @@ func strNew(f *Frame, t *Type, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			n--
+		}
+		for i := 0; i < len(s); i++ {
+			buf.WriteByte(s[i])
+			if n > 0 {
+				buf.WriteString(sub)
+				n--
+			} else {
+				i++
+				if i < len(s) {
```
trotterdylan: Can we do this branch after the loop? I think the code would read more clearly: do the required replacements, then write the remainder of the string.

I think things would be tidier if your loop condition dictated when you exit the loop:

```
i := 0
for n > 0 && i < numBytes {
  // .. replacement logic
  i++
  n--
}
```

As written it's not immediately obvious what the termination condition is.

runtime/str.go
S-YOU: `"".replace("", "x") returns "x", but `"".replace("", "x",1)` return "" in CPython, so it is need imo.

runtime/str.go
S-YOU: I'll try.

runtime/str.go
S-YOU: Great idea, thank you.

runtime/str.go

S-YOU:
Updated those, Please take a look again.

PR #170, opened 2017-01-19
Minor changes to make CPython sequence tests pass

trotterdylan:
Also updated Makefile to include these tests in the "make test" suite.
It became awkward to automatically detect the .py test files (they don't
really follow any particular naming convention) so they're listed
explicitly. To make things easier to follow I also listed out the other
stdlib test files explicitly. Long term I think we should be listing all
files this way (including .go sources) so that things are less magical.


trotterdylan:
@S-YOU This PR activates the stdlib tests that you've checked in as part of the normal test suite. It was made possible by all your hard work :)

S-YOU:
Wow, thank you very much to including in.

PR #176, opened 2017-01-20
Implement string interpolation conversion type 'o'

choleraehyq:
Conversion type 'o' represents signed octal value.

Signed-off-by: Cholerae Hu <choleraehyq@gmail.com>


runtime/str.go
```diff
@@ -753,6 +753,12 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
 					return nil, raised
 				}
 				val = s.Value()
+			} else if matches[7] == "o" {
+				if o.isInstance(LongType) {
+					val = toLongUnsafe(o).Value().Text(8)
```
trotterdylan: I realized the other day when I was looking at this code that CPython uses the __hex__ and __oct__ slots for this: https://github.com/python/cpython/blob/2.7/Objects/stringobject.c#L4019

So I think the right thing to do here is to implement those slots for int and long types. I think it would make this code considerably simpler.

choleraehyq: Totally agreed. I'll work on this.

choleraehyq: @trotterdylan The behavior is weird.
```
class A(object):
  def __oct__(self):
    return '0123'

class B(object):
  def __oct__(self):
    return '0124'
  def __int__(self):
    return 8

# This line will cause a TypeError
# TypeError: %o format: a number is required, not A
# print '%o' % A()

# This line is correct
print '%o' % B()
```
Obviously this conversion doesn't use `__hex__` and `__oct__` now.

trotterdylan: Huh, that is very weird. It appears that CPython does not call `__hex__`/`__oct__`. So I'm confused about what the code in stringobject.c is actually doing.

trotterdylan: So I guess your implementation is consistent with CPython and we should leave it as is. Will approve and merge.

Issue #72, opened 2017-01-09
Support decorators PEP 318


trotterdylan:
https://www.python.org/dev/peps/pep-0318/

trotterdylan:
Note that for starters it'd be nice if we would just fail hard when decorators are present in the ast.

S-YOU:
@trotterdylan 
I am trying to implement to fail hard, but I am seeing alot of decorators already in the code

https://github.com/google/grumpy/blob/master/third_party/stdlib/_abcoll.py
third_party/stdlib/UserDict.py
third_party/stdlib/collections.py

trotterdylan:
We can replace those with the verbose 'foo = decorator(foo)' for the time
being.

S-YOU:
> We can replace those with the verbose 'foo = decorator(foo)' for the time
being.

Is that mean [classmethod](#73) has to be implemented first?

trotterdylan:
Yeah I guess that is necessary first :\

PR #177, opened 2017-01-20
Implement Decorators

S-YOU:
This is for #72, but I guess this implementation may be more a like a hack, because Instead of generating Go code, I triggered AST to add more nodes while walking through.

- Confirmed simple decorators working, not sure there is some complex use cases.
- This is only for function decorators, class decorators not yet supported.
(I don't get this syntax for class working yet. `C = decorator(C)`)

```
% make run
def bold(fn):
    def wrapped():
        return "<b>" + fn() + "</b>"
    return wrapped

def italic(fn):
    def wrapped():
        return "<i>" + fn() + "</i>"
    return wrapped

@bold
@italic
def hello():
    return "hello world"

print hello()
```
```
<b><i>hello world</i></b>
```

```
import contextlib

@contextlib.contextmanager
def tag(name):
    print "<%s>" % name
    yield
    print "</%s>" % name

with tag("h1"):
  print "foo"
```
```
<h1>
foo
</h1>
```

### DIFF
--- a/compiler/stmt_test.py
+++ b/compiler/stmt_test.py
@@ -212,6 +212,17 @@ class StatementVisitorTest(unittest.TestCase):
         util.ParseError, "'continue' not in loop",
         _ParseAndVisit, 'for i in (1,):\n  pass\nelse:\n  continue')
 
+  def testFunctionDecorator(self):
+    self.assertEqual((0, '<b>foo</b>\n'), _GrumpRun(textwrap.dedent("""\
+        def bold(fn):
+          def wrapped():
+            return '<b>' + fn() + '</b>'
+          return wrapped
+        @bold
+        def foo():
+          return 'foo'
+        print foo()""")))
+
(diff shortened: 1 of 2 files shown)


compiler/stmt_test.py
```diff
@@ -212,6 +212,17 @@ def testForElseContinueNotNested(self):
         util.ParseError, "'continue' not in loop",
         _ParseAndVisit, 'for i in (1,):\n  pass\nelse:\n  continue')
 
+  def testFunctionDecorator(self):
+    self.assertEqual((0, '<b>foo</b>\n'), _GrumpRun(textwrap.dedent("""\
+        def bold(fn):
+          def wrapped():
```
trotterdylan: slightly shorter:

```
def bold(fn):
  return lambda: '<b>' + fn() + '</b>'
```

S-YOU:
Sorry for late update, updated those.

PR #182, opened 2017-01-21
Fix str split with whitespace

S-YOU:
This should fix following cases.
```
assert "".split() == []
assert " ".split() == []
assert " a ".split() == ['a']
assert 'a b c d '.split() == ['a', 'b', 'c', 'd']
assert ' a b c d '.split(None, 1) == ['a', 'b c d ']
assert '   a b c d'.split(None, 0) == ['a b c d']
```
Just not sure my approach is correct, that's currently I had in mind.


runtime/str.go
```diff
@@ -590,7 +590,14 @@ func strSplit(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	s := toStrUnsafe(args[0]).Value()
 	var parts []string
 	if sep == "" {
+		s = strings.TrimLeft(s, string(strASCIISpaces))
 		parts = whitespaceSplitRegexp.Split(s, maxSplit)
+		l := len(parts)
+		if l > 0 {
+			if strings.Trim(parts[l-1], string(strASCIISpaces)) == "" {
```
trotterdylan: Combine this condition with the if statement above: if l > 0 && ...

PR #185, opened 2017-01-21
[WIP] Add stringformat

S-YOU:
- Still finding way to monkey patch properly.
- Temporarily put the file in pypy, please do not merge.

```
% make run
import stringformat
print ['{a:5} {} {:d} {:<10}'.format("abcd", 1234, "a", a=42)]
['   42 abcd 1234 a         ']
```

### DIFF
--- a/third_party/pypy/stringformat.py
+++ b/third_party/pypy/stringformat.py
@@ -1,0 +1,399 @@
+# -*- coding: utf-8 -*-
+"""Advanced string formatting for Python >= 2.4.
+
+An implementation of the advanced string formatting (PEP 3101).
+
+Author: Florent Xicluna
+"""
+
+import re
+
+__all__ = ['FormattableString', 'init']
+
+def isdigit(s):
+  for x in s:
+    if x < "0" or x > "9":
+      return False
+  return True
+
+def setdefault(d, k, v):
+  if k not in d:
+    d[k] = v
+  return d[k]
+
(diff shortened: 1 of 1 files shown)


trotterdylan:
This is cool. I can think of a couple different options for integrating this off the top of my head:

1. Add some code to the code generator that imports stringformat automatically as part of main()
2. Support a code generator mode where code will assume it's part of the core grumpy package and include the generated code directly into the runtime build

The nice thing about 2. is that we could do something similar for the `__builtin__` module so that we could implement some builtins in Python as has been discussed in some PRs.

The downside is that the code generator gets more complicated and it needs to hardcode a small set of imports that it will pull from sys.modules and not import the associated Go package.

S-YOU:
I guess with `1.`, we cannot use that feature internally, like in third_party modules, unless manually importing it.

trotterdylan:
I was thinking the code would run before any other Python code, so it should work for third_party modules. But 1. does seem a lot more hacky 2.

There are a couple parts of 2. that would need to be implemented:

a. Special casing some modules in the compiler so that they're pulled directly from sys.modules but no corresponding Go package is imported
b. Supporting a code generation mode where the module's Code object is called something different and grumpy API calls are not prefixed with `πg.`

I'm still not convinced this is the best approach (there may be others we haven't thought of) but it does seem doable.

PR #186, opened 2017-01-21
Make `str` casing methods multi-byte aware

meadori:
In the course of reviewing #116 it came to light that
some of the other `str` methods that involve changing
case do not handle multi-byte characters correctly.
This patch fixes that.


runtime/str.go
```diff
@@ -157,29 +158,35 @@ func strAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return NewStr(stringV + stringW).ToObject(), nil
 }
 
+func toLower(b byte) byte {
```
trotterdylan: Please group helper functions to the bottom of the file.

meadori: Fixed.  Thanks for the review!

PR #188, opened 2017-01-21
Add bm_call_method and bm_call_simple from the python benchmark suite.

nairb774:
Minimal change were made - mostly a change from using the perf module to using weetest. This helps a little with gauging where grumpy is when compared to CPython.

### DIFF
--- a/benchmarks/bm_call_simple.py
+++ b/benchmarks/bm_call_simple.py
@@ -1,0 +1,139 @@
+"""Microbenchmark for function call overhead.
+
+This measures simple function calls that are not methods, do not use varargs or
+kwargs, and do not use tuple unpacking.
+
+Taken from:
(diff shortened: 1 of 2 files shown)


benchmarks/bm_call_simple.py
```diff
@@ -0,0 +1,139 @@
...
+Taken from:
+https://github.com/python/performance/blob/9b8d859/performance/benchmarks/bm_call_simple.py
+"""
+
+import weetest
+
+
+def foo(a, b, c, d):
+  # 20 calls
+  bar(a, b, c)
```
trotterdylan: If you want to avoid capturing some of the overhead of the globals lookup you could store bar as a local. But maybe it's useful to have that work factored in.

nairb774: I wouldn't make any changes here - this is something that I would either want to capture upstream, or improve in grumpy.

PR #189, opened 2017-01-21
Allow passing Python functions to Go for invocation.

nairb774:
Big WIP - This code is gnarly and rather than continuing to sit on it, thoughts on how to handle all the nasty `panics` below? There has to be better ways. Regardless, this opens up a lot more interoperability with the Go ecosystem - just a matter of how to handle error cases.

### DIFF
--- a/runtime/native.go
+++ b/runtime/native.go
@@ -479,22 +479,95 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
 	if raised != nil {
 		return reflect.Value{}, raised
 	}
-	rtype := val.Type()
 	for {
+		rtype := val.Type()
 		if rtype == expectedRType {
 			return val, nil
 		}
 		if rtype.ConvertibleTo(expectedRType) {
 			return val.Convert(expectedRType), nil
 		}
-		if rtype.Kind() == reflect.Ptr {
+		switch rtype.Kind() {
+		case reflect.Ptr:
 			val = val.Elem()
-			rtype = val.Type()
 			continue
+
+		case reflect.Func:
+			if fn, ok := val.Interface().(func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)); ok {
+				val = nativeToPyFuncBridge(Func(fn), expectedRType)
+				continue
+			}
 		}
-		break
+		return val, f.RaiseType(TypeErrorType, fmt.Sprintf("cannot convert %s to %s", rtype, expectedRType))
 	}
-	return reflect.Value{}, f.RaiseType(TypeErrorType, fmt.Sprintf("cannot convert %s to %s", rtype, expectedRType))
+}
+
+func nativeToPyFuncBridge(fn Func, target reflect.Type) reflect.Value {
+	firstInIsFrame := target.NumIn() > 0 && target.In(0) == reflect.TypeOf((*Frame)(nil))
+
+	outs := make([]reflect.Type, target.NumOut())
+	for i := range outs {
+		outs[i] = target.Out(i)
+	}
+
+	return reflect.MakeFunc(target, func(args []reflect.Value) []reflect.Value {
+		var frame *Frame
+		if firstInIsFrame {
+			frame, args = args[0].Interface().(*Frame), args[1:]
+		} else {
+			frame = NewRootFrame()
+		}
+
+		pyArgs := frame.MakeArgs(len(args))
+		for i, arg := range args {
+			var raised *BaseException
+			pyArgs[i], raised = WrapNative(frame, arg)
+			if raised != nil {
+				panic(fmt.Sprintf("WrapNative(%v)=%v", arg, raised)) // TODO: Figure this out...
+			}
+		}
+
+		ret, raised := fn(frame, pyArgs, nil)
+		if raised != nil {
+			panic(fmt.Sprintf("python error %v crossing into Go", raised)) // TODO: Figure this out...
+		}
+
+		switch len(outs) {
+		case 0:
+			if ret != nil && ret != None {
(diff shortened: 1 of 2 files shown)


runtime/native.go
```diff
@@ -479,22 +479,95 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
+		} else {
+			frame = NewRootFrame()
+		}
+
+		pyArgs := frame.MakeArgs(len(args))
+		for i, arg := range args {
+			var raised *BaseException
+			pyArgs[i], raised = WrapNative(frame, arg)
+			if raised != nil {
+				panic(fmt.Sprintf("WrapNative(%v)=%v", arg, raised)) // TODO: Figure this out...
```
trotterdylan: I don't have any great suggestions here. Given that we can't assume anything about the nature or the context of the call, panic is probably the only general option available to us. In the specific case where the last return value is *BaseException, we could return the raised exception, but that's going to have limited applicability.

The value panicked could just be the raised exception, that way at least there's some opportunity to catch it and recover.

runtime/native.go
```diff
@@ -479,22 +479,95 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
+
+		case 1:
+			v, raised := maybeConvertValue(frame, ret, outs[0])
+			if raised != nil {
+				panic(fmt.Sprintf("maybeConvertValue(%v)=%v", ret, raised)) // TODO: Figure this out...
+			}
+			return []reflect.Value{v}
+		}
+
+		// TODO: Support other iterables?
```
trotterdylan: This should be pretty easy to accomplish with seqForEach() or similar.

runtime/native.go
```diff
@@ -479,22 +479,95 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
 		}
-		if rtype.Kind() == reflect.Ptr {
+		switch rtype.Kind() {
+		case reflect.Ptr:
 			val = val.Elem()
-			rtype = val.Type()
 			continue
+
+		case reflect.Func:
+			if fn, ok := val.Interface().(func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)); ok {
```
nairb774: The common case should be that when `value.Kind()==Func` this interface assertion should pass (at least today). If other built in types were returning other func like values from their Native slot, then this might need to be adjusted.

runtime/native.go
```diff
@@ -479,22 +479,95 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
+				val = nativeToPyFuncBridge(Func(fn), expectedRType)
```
nairb774: Kind of, the function type of `nativeToPyFuncBridge` was wrong. This was left from a bit of debugging/re-writing I did earlier.

runtime/native.go
```diff
@@ -479,22 +479,95 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
+				continue
+			}
 		}
-		break
+		return val, f.RaiseType(TypeErrorType, fmt.Sprintf("cannot convert %s to %s", rtype, expectedRType))
+	}
+}
+
+func nativeToPyFuncBridge(fn Func, target reflect.Type) reflect.Value {
+	firstInIsFrame := target.NumIn() > 0 && target.In(0) == reflect.TypeOf((*Frame)(nil))
```
nairb774: Done.

runtime/native.go
nairb774: Works for me.

runtime/native.go
nairb774: PTAL.

nairb774:
This still needs some tests. I'll start working on those now that this looks like it is in a better place, but feel free to review the code in the implementation.

runtime/native.go
```diff
@@ -479,22 +479,115 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
+				val = nativeToPyFuncBridge(fn, expectedRType)
+				continue
+			}
 		}
-		break
+		return val, f.RaiseType(TypeErrorType, fmt.Sprintf("cannot convert %s to %s", rtype, expectedRType))
+	}
+}
+
+var baseExceptionReflectType = reflect.TypeOf((*BaseException)(nil))
```
trotterdylan: Group these globals at the top of the file please.

nairb774: Done.

nairb774:
PTAL.

runtime/native.go
```diff
@@ -489,22 +492,111 @@ func maybeConvertValue(f *Frame, o *Object, expectedRType reflect.Type) (reflect
...
 		}
-		if rtype.Kind() == reflect.Ptr {
+		switch rtype.Kind() {
+		case reflect.Ptr:
 			val = val.Elem()
-			rtype = val.Type()
 			continue
+
+		case reflect.Func:
+			if fn, ok := val.Interface().(func(*Frame, Args, KWArgs) (*Object, *BaseException)); ok {
```
trotterdylan: This strikes me as not quite right because it only (currently) really works for function objects which return their Call method (IIUC). I wonder if we should be checking whether `expectedRType.Kind() == reflect.Func` upfront and if so, checking whether `o.__call__` exists. If so then produce a bridge for it, otherwise, fall back to ToNative.

runtime/native_test.go
```diff
@@ -422,6 +422,122 @@ func TestMaybeConvertValue(t *testing.T) {
 	}
 }
 
+func TestNativveToPyFuncBridge(t *testing.T) {
```
trotterdylan: `s/TestNativve/TestNative/`

PR #192, opened 2017-01-22
[WIP] Implement members / wildcard imports

S-YOU:
- I think this is more like a PoC, by trying to `__import__` module_name, if it does not work assume member imports, which means modules must be available at compile time.
- Wildcard imports only work for pure python modules, without `from __go__` statements on any of the dependencies. 
- Wildcard imports only look in $PWD/third_party/stdlib, $PWD/third_party/pypy, which is hard-coded, need to find better way, also using getcwd() does not look right.

```
% make run 
import os
import os.path
from os import path
from os import chdir
from os import path, chdir
from os import path as os_path, chdir as cd
from os.path import abspath, join as j
from test import test_support, seq_tests
from test import test_support
from test import test_support as support
import test.test_support
from test import mapping_tests

from sre_constants import ANY, BRANCH
from StringIO import *

print globals().keys()

^D
['test_support', 'chdir', 'StringIO', 'support', '__file__', 'j', 'cd', 'mapping_tests', 'sre_constants', 'ANY', 'seq_tests', 'BRANCH', 'path', 'test', '__name__', 'os_path', 'os', 'abspath', 'os.path']
```

If approach is acceptable I could continue to add tests and chore a bit more.

### DIFF
--- a/compiler/stmt.py
+++ b/compiler/stmt.py
@@ -16,6 +16,8 @@
 
 """Visitor class for traversing Python statements."""
 
+import os
+import sys
 import ast
 import string
 import textwrap
@@ -375,12 +377,6 @@ class StatementVisitor(ast.NodeVisitor):
         self.block.bind_var(self.writer, asname, mod.expr)
 
   def visit_ImportFrom(self, node):
-    # Wildcard imports are not yet supported.
-    for alias in node.names:
-      if alias.name == '*':
-        msg = 'wildcard member import is not implemented: from %s import %s' % (
-            node.module, alias.name)
-        raise util.ParseError(node, msg)
     self._write_py_context(node.lineno)
     if node.module.startswith(_NATIVE_MODULE_PREFIX):
       values = [alias.name for alias in node.names]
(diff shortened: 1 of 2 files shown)


S-YOU:
Looks like some python installation don't have test files in test directory, which breaking this. Looking into it.

nairb774:
Why do it at compile time and not at runtime?

trotterdylan:
I agree, I would do the wildcard expansion at runtime. You can just import the module as normal and put it in a temporary, then generate code that attempts to get `__all__` or fall back to getting all keys in the module dict.

Also, I suggest breaking the wildcard support apart from directly importing module members. I think the latter is going to require grumpc to have knowledge of the package path. The implementation you have is clever but it requires CPython and long term I'd like Grumpy to be able to build and run grumpc ("self-hosting").

Once grumpc is self-hosting, what I have in mind is that grumpc could use the `__go__` extensions to access the the [go.build](https://golang.org/pkg/go/build/) package to inspect the GOPATH to learn what packages exist. Knowing that, it can decide whether the import statement is referencing a module or members of that module.

Given how much work you've put into getting the standard libraries working, I would say we're not all that far off from being able to make grumpc self-hosting. The biggest hurdle now seems to be the ast module. The meat of it is unfortunately not pure python in either PyPy or CPython. I have heard rumors there is a project to implement it in Python but I couldn't find it from a quick Google search.

S-YOU:
I completely forgot that those could be done at runtime, :D It was the idea come up when I hacking around compiler.

Yeah, I have a hope for running grumpy self hosting too. Noticed you havent included Cpython specific module in compiler. Adding os, sys to that is quite dirty too.

I will close this soon, unless you need some part of it.

trotterdylan:
Closing sounds fine to me. Thanks for digging into this!

trotterdylan:
The [PythonParser project](https://github.com/m-labs/pythonparser) may be a viable pure-Python alternative to the ast module.

PR #193, opened 2017-01-22
Add builtin setattr

S-YOU:
- Added few tests in python side, but go side still have to figure out.


S-YOU:
To write go tests, I should write a helper function that use SetAttr and also use GetAttr to return value from it?

trotterdylan:
Yeah that sounds like a good approach.

S-YOU:
Added separate test function, because I can't find a way to put it in `TestBuiltinFuncs`

PR #194, opened 2017-01-22
Add stdlib list, slice tests

S-YOU:
- Added stdlib list and slice tests
- Added those to Makefile
- Removed seq_tests, mapping_tests, `*_tests` from Makefile, there are just dependent by `test_*.py`, since they don't actually test themselves.
- minor chore on `unittest/__init__.py`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,11 +78,10 @@ STDLIB_TESTS := \
   re_tests \
   sys_test \
   tempfile_test \
-  test/seq_tests \
   test/test_tuple \
-  test/mapping_tests \
   test/test_dict \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -78,11 +78,10 @@ STDLIB_TESTS := \
   re_tests \
   sys_test \
   tempfile_test \
-  test/seq_tests \
```
trotterdylan: Were these removals intentional?

S-YOU: They exported classes for actual tests file like test_tuple, test_list, but they dont do actual test themselves, take a look at one of those.

PR #198, opened 2017-01-23
Implement the sequence `count` method

meadori:
As specified here:

* https://docs.python.org/2/library/stdtypes.html#sequence-types-str-unicode-list-tuple-bytearray-buffer-xrange

This commit only uses the sequence count method for `list.count`,
but it can easily be applied to other sequence types in follow-on
commits.

### DIFF
--- a/runtime/seq_test.go
+++ b/runtime/seq_test.go
@@ -43,6 +43,38 @@ func TestSeqApply(t *testing.T) {
 	}
 }
 
+func TestSeqCount(t *testing.T) {
+	badEqType := newTestClass("Eq", []*Type{IntType}, newStringDict(map[string]*Object{
+		"__eq__": newBuiltinFunction("__eq__", func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+			return nil, f.RaiseType(TypeErrorType, "uh oh")
+		}).ToObject(),
+	}))
+	badNonZeroType := newTestClass("BadNonZeroType", []*Type{ObjectType}, newStringDict(map[string]*Object{
+		"__nonzero__": newBuiltinFunction("__nonzero__", func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+			return nil, f.RaiseType(TypeErrorType, "uh oh")
+		}).ToObject(),
+	}))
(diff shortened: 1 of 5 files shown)


runtime/seq_test.go
```diff
@@ -43,6 +43,43 @@ func TestSeqApply(t *testing.T) {
 	}
 }
 
+func TestSeqCount(t *testing.T) {
+	fun := wrapFuncForTest(func(f *Frame, seq *Object, v *Object) (*Object, *BaseException) {
```
trotterdylan: You could wrap seqCount directly instead of creating a new function: wrapFuncForTest(seqCount)

runtime/seq_test.go
```diff
@@ -43,6 +43,43 @@ func TestSeqApply(t *testing.T) {
...
+	worseEqType := newTestClass("WorseEqCmp", []*Type{IntType}, newStringDict(map[string]*Object{
+		"__eq__": newBuiltinFunction("__eq__", func(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+			return newObject(badNonZeroType), nil
+		}).ToObject(),
+	}))
+	cases := []invokeTestCase{
+		{args: wrapArgs(newTestList(), NewInt(1)), want: NewInt(0).ToObject()},
+		{args: wrapArgs(newTestList(1, 2, 3, 4, 5, 6), NewInt(7)), want: NewInt(0).ToObject()},
+		{args: wrapArgs(newTestList(1, 2, 3, 2, 2, 4), NewInt(2)), want: NewInt(3).ToObject()},
+		{args: wrapArgs(newTestList(newObject(badEqType)), newObject(badEqType)), wantExc: mustCreateException(TypeErrorType, "uh oh")},
```
trotterdylan: Test some other sequence types besides list

runtime/seq_test.go
meadori: Ah, that is much nicer.  Thanks!

runtime/seq_test.go
meadori: Will Do.  Thanks for the review!

meadori:
The latest commit should have addressed all the review feedback.

PR #201, opened 2017-01-24
Add a bunch of micro benchmarks.

nairb774:
Method covered by new benchmarks:
* Dict.GetItem
* Dict.IterItems
* Dict.IterKeys
* Dict.IterValues
* NewInt - Fixed so that references aren't elided anymore.
* NewStr

I have some improvements in the Dict area and I thought it might be good to get some benchmarks on the ground first.

### DIFF
--- a/runtime/dict_test.go
+++ b/runtime/dict_test.go
@@ -17,6 +17,7 @@ package grumpy
 import (
 	"reflect"
 	"regexp"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


nairb774:
The keep alive (currently) helps prevent the Go compiler from removing the assignment to the variables and getting too clever. In the one case, (NewInt) was basically being inlined away. While I am sure this isn't a correct solution for the long term, it seems to work today. If for some reason a new Go version makes the benchmarks much much faster, then we have some work to do :)

PR #202, opened 2017-01-24
Add builtin delattr

S-YOU:


runtime/builtin_types_test.go
```diff
@@ -23,6 +23,33 @@ import (
 	"testing"
 )
 
+func TestBuiltinDelAttr(t *testing.T) {
+	fun := wrapFuncForTest(func(f *Frame, args ...*Object) (*Object, *BaseException) {
+		fooObject := newTestClass("Foo", []*Type{ObjectType}, toDictUnsafe(args[0])).ToObject()
```
trotterdylan: Rather than creating a test class for each test case, I would create a single test class and create objects of that type that you create inside TestBuiltinDelAttr. Classes are relatively expensive to create and I think we mostly want to test normal object delattr rather than delattr on types.

I just realized you did something similar in SetAttr (sorry for not noticing) but I think we should change that one too. That doesn't need to happen in this PR though.

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,33 @@ import (
 	"testing"
 )
 
+func TestBuiltinDelAttr(t *testing.T) {
+	fun := wrapFuncForTest(func(f *Frame, args ...*Object) (*Object, *BaseException) {
```
trotterdylan: This can be `func(*Frame, o *Object, args ...*Object) (*Object, *BaseException)`. Then you don't have to mess around with args slices below.

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,33 @@ import (
...
+		fooObject := newTestClass("Foo", []*Type{ObjectType}, toDictUnsafe(args[0])).ToObject()
+		result, raised := builtinDelAttr(f, append(Args{fooObject}, args[1:]...), nil)
```
trotterdylan: I would fetch "delattr" from the builtins dict and call it, rather than calling the underlying function directly. That will exercise the dispatch logic and ensure that the "delattr" key is in builtins.

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,33 @@ import (
...
+		if raised != nil || result != None {
+			return result, raised
+		}
+		result, raised = GetAttr(f, fooObject, toStrUnsafe(args[1]), nil)
+		if raised != nil || result != nil {
+			return None, raised
```
trotterdylan: I think it's preferable to return an exception only when the first return value is nil. Therefore, I would just `return GetAttr(...)`. That will be simpler anyway.

testing/builtin_test.py
```diff
@@ -247,14 +247,44 @@ def __init__(self, x):
...
 
 class Foo(object):
   pass
 
 setattr(Foo, "a", 1)
 assert Foo.a == 1  # pylint: disable=no-member
 
+delattr(Foo, "a")
+try:
+  getattr(Foo, "a")
```
trotterdylan: Slightly less verbose to use None as a special marker value:

```
assert getattr(Foo, "a", None) is None
```

runtime/builtin_types_test.go
S-YOU: If I create objects inside TestBuiltinDelAttr, I thini delAttr or setAttr on one test may modify the object and it will effect another tests. Tests have to run in particular order for that case, I guess.
Is that acceptable?

runtime/builtin_types_test.go
S-YOU: Sure, I will change to use it.

runtime/builtin_types_test.go
trotterdylan: Just create the class in TestBuiltinDelAttr and then instantiate objects of that class in the args for each test case: `wrapArgs(newObject(fooType), ...)`

runtime/builtin_types_test.go
S-YOU: Thanks, I complicated things much.

testing/builtin_test.py
S-YOU: Sure

runtime/builtin_types_test.go
S-YOU: Oh, ok. Thank you

S-YOU: It turns out that I also need to do SetAttr to confirm DelAttr, so I need to pass newTestDict() and SetAttr inside inner functions. Any better idea do you have?

S-YOU: I just found out actual tests in TestObjectDelAttr in object_test.go. It does not seems to dynamically  SetAttr first for each tests, so may be I will follow that way.

runtime/builtin_types_test.go
S-YOU: Thank you, I didn't know that.

S-YOU:
Updated those, please take a look again. Thanks.

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,32 @@ import (
 	"testing"
 )
 
+func TestBuiltinDelAttr(t *testing.T) {
+	f := NewRootFrame()
+	delattr := mustNotRaise(Builtins.GetItemString(NewRootFrame(), "delattr"))
```
trotterdylan: You can use `f` instead of NewRootFrame() here

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,32 @@ import (
...
+	fooType := newTestClass("Foo", []*Type{ObjectType}, newStringDict(map[string]*Object{}))
+	fooForDelAttr := newObject(fooType)
+	SetAttr(f, fooForDelAttr, NewStr("bar"), None)
```
trotterdylan: You should check that SetAttr doesn't raise. You can do:

```
mustNotRaise(nil, SetAttr(f, fooForDelAttr, NewStr("bar"), None))
```

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,32 @@ import (
...
+	fun := wrapFuncForTest(func(f *Frame, args ...*Object) (*Object, *BaseException) {
+		_, raised := delattr.Call(f, args, nil)
+		if raised != nil {
+			return nil, raised
+		}
+		return GetAttr(f, args[0], toStrUnsafe(args[1]), nil)
```
trotterdylan: I would use None (or something else) as a marker value so that GetAttr doesn't raise in the successful case. That will make the test cases a little more easy to understand (took me a minute to understand why all the test cases had wantExc).

I would also return the result of delattr along with the GetAttr result in a tuple (like in TestBuiltinSetAttr below) so that you can verify that delattr returns None.

runtime/builtin_types_test.go
```diff
@@ -23,6 +23,32 @@ import (
 	"testing"
 )
 
+func TestBuiltinDelAttr(t *testing.T) {
+	f := NewRootFrame()
+	delattr := mustNotRaise(Builtins.GetItemString(NewRootFrame(), "delattr"))
+	fooType := newTestClass("Foo", []*Type{ObjectType}, newStringDict(map[string]*Object{}))
```
S-YOU: nil panic, but NewDict() works, Thank you.

runtime/builtin_types_test.go
S-YOU: Sure

runtime/builtin_types_test.go
S-YOU: I am getting strange results, I lost somewhere. please take a look (For first test case).
It is throwing, even if I return nil in exception.

```
% git diff                                                                                                                     [add_delattr]$
diff --git a/runtime/builtin_types_test.go b/runtime/builtin_types_test.go
index 53eeca8..f1f3aee 100644
--- a/runtime/builtin_types_test.go
+++ b/runtime/builtin_types_test.go
@@ -26,18 +26,24 @@ import (
 func TestBuiltinDelAttr(t *testing.T) {
        f := NewRootFrame()
        delattr := mustNotRaise(Builtins.GetItemString(NewRootFrame(), "delattr"))
-       fooType := newTestClass("Foo", []*Type{ObjectType}, newStringDict(map[string]*Object{}))
+       fooType := newTestClass("Foo", []*Type{ObjectType}, NewDict())
        fooForDelAttr := newObject(fooType)
-       SetAttr(f, fooForDelAttr, NewStr("bar"), None)
+       mustNotRaise(nil, SetAttr(f, fooForDelAttr, NewStr("bar"), None))
        fun := wrapFuncForTest(func(f *Frame, args ...*Object) (*Object, *BaseException) {
-               _, raised := delattr.Call(f, args, nil)
+               result, raised := delattr.Call(f, args, nil)
                if raised != nil {
                        return nil, raised
                }
-               return GetAttr(f, args[0], toStrUnsafe(args[1]), nil)
+               val, raised := GetAttr(f, args[0], toStrUnsafe(args[1]), nil)
+               if raised != nil {
+                       fmt.Println(22222222, args, result, val == nil, raised)
+                       return newTestTuple(result, val == nil).ToObject(), nil
+               }
+               fmt.Println(333333333, args)
+               return nil, f.RaiseType(AttributeErrorType, fmt.Sprintf("'delattr' failed to remove '%s', got '%s' instead", args[1], val))
        })
        cases := []invokeTestCase{
-               {args: wrapArgs(fooForDelAttr, "bar"), wantExc: mustCreateException(AttributeErrorType, "'Foo' object has no attribute 'bar'")},
+               {args: wrapArgs(fooForDelAttr, "bar"), want: newTestTuple(None, True.ToObject()).ToObject()},
                {args: wrapArgs(fooForDelAttr, "baz"), wantExc: mustCreateException(AttributeErrorType, "'Foo' object has no attribute 'baz'")},
                {args: wrapArgs(fooForDelAttr), wantExc: mustCreateException(TypeErrorType, "'delattr' requires 2 arguments")},
                {args: wrapArgs(fooForDelAttr, "foo", "bar"), wantExc: mustCreateException(TypeErrorType, "'delattr' requires 2 arguments")},

/Users/name/repo/grumpy% make build/runtime.pass
22222222 [<Foo object at 0xc420425f20> 'bar'] None true AttributeError('\'Foo\' object has no attribute \'bar\'',)
--- FAIL: TestBuiltinDelAttr (0.00s)
	builtin_types_test.go:53: func1(<Foo object at 0xc420425f20>, 'bar') raised AttributeError('\'Foo\' object has no attribute \'bar\'',), want <nil>
Exception: foo
RuntimeError: foo
FAIL
FAIL	grumpy	1.197s
make: *** [build/runtime.pass] Error 1
```

runtime/builtin_types_test.go
S-YOU: That is needed for `SetAttr(f, fooForDelAttr, NewStr("bar"), None)`,
Am I missing something?

trotterdylan: `Builtins.GetItemString(NewRootFrame(), ...)` can be `Builtins.GetItemString(f, ...)`. You can use a frame more than once. If it raises then you need to `f.RestoreExc(nil, nil)`, but mustNotRaise() ensures this is not raising.

S-YOU: Ah, sorry, I misread wrong line.

runtime/builtin_types_test.go
S-YOU: Looks like I needed a f.RestoreExc(nil, nil) here too.

trotterdylan: Yeah nativeInvoke() in native.go (which is used indirectly by wrapFuncForTest) [checks f.ExcInfo()](https://github.com/google/grumpy/blob/master/runtime/native.go#L562) when returning from invoking the function.

But you could instead just provide None instead of nil as the last arg to GetAttr() and then it'll return None instead of raising AttributeError.

S-YOU:
Updated those, thanks for iterations.

runtime/builtin_types_test.go
S-YOU: Good idea. Will update.

S-YOU: Hmm, I set None to bar, and I guess I cannot differentiate that None, and default None that way?

trotterdylan: Oh right. You could use some other value to differentiate. The most foolproof way is to create a newObject(ObjectType) which will only ever equal itself.

S-YOU: Something like this? 
```
% git diff                                                                                                                          [master]$
diff --git a/runtime/builtin_types_test.go b/runtime/builtin_types_test.go
index 6ba50bb..d5120d7 100644
--- a/runtime/builtin_types_test.go
+++ b/runtime/builtin_types_test.go
@@ -28,21 +28,18 @@ func TestBuiltinDelAttr(t *testing.T) {
        delattr := mustNotRaise(Builtins.GetItemString(f, "delattr"))
        fooType := newTestClass("Foo", []*Type{ObjectType}, NewDict())
        fooForDelAttr := newObject(fooType)
-       mustNotRaise(nil, SetAttr(f, fooForDelAttr, NewStr("bar"), None))
+       fooValue := newObject(ObjectType)
+       mustNotRaise(nil, SetAttr(f, fooForDelAttr, NewStr("bar"), fooValue))
        fun := wrapFuncForTest(func(f *Frame, args ...*Object) (*Object, *BaseException) {
                result, raised := delattr.Call(f, args, nil)
                if raised != nil {
                        return nil, raised
                }
-               val, raised := GetAttr(f, args[0], toStrUnsafe(args[1]), nil)
-               if raised != nil && raised.isInstance(AttributeErrorType) {
-                       f.RestoreExc(nil, nil)
-                       return newTestTuple(result, val == nil).ToObject(), nil
-               }
-               return nil, f.RaiseType(AttributeErrorType, fmt.Sprintf("'delattr' failed to remove '%s', got '%s' instead", args[1], val))
+               val, raised := GetAttr(f, args[0], toStrUnsafe(args[1]), None)
+               return newTestTuple(result, val == fooValue).ToObject(), nil
        })
        cases := []invokeTestCase{
-               {args: wrapArgs(fooForDelAttr, "bar"), want: newTestTuple(None, True.ToObject()).ToObject()},
+               {args: wrapArgs(fooForDelAttr, "bar"), want: newTestTuple(None, False.ToObject()).ToObject()},
                {args: wrapArgs(fooForDelAttr, "baz"), wantExc: mustCreateException(AttributeErrorType, "'Foo' object has no attribute 'baz'")},
                {args: wrapArgs(fooForDelAttr), wantExc: mustCreateException(TypeErrorType, "'delattr' requires 2 arguments")},
                {args: wrapArgs(fooForDelAttr, "foo", "bar"), wantExc: mustCreateException(TypeErrorType, "'delattr' requires 2 arguments")},
```

Should I create new PR?

trotterdylan: Yeah that looks nicer. I would just return val rather than returning val == fooValue and you can pass fooValue in the want: value.

S-YOU: Oh, you're right.

S-YOU: fooValue will not return because deleted, instead it return None.

PR #205, opened 2017-01-24
Add set literal

aisk:


aisk:
I signed it!

compiler/expr_visitor_test.py
```diff
@@ -35,11 +35,13 @@ def Test(self):
   return Test
 
 
-def _MakeLiteralTest(lit):
+def _MakeLiteralTest(lit, expected=None):
   def Test(self):
     status, output = _GrumpRun('print repr({!r}),'.format(lit))
```
trotterdylan: Hmm this is bit of a weird case because the result of repr() on a set does not give back the same literal string that produced it, so I see why you're changing this. But I don't think it's working quite the way we want. In this case Grumpy is running the statement `print repr(set(['foo', 'bar']))`. It would be better if it was running `print repr({'foo', 'bar'})`. Maybe the best thing to do is for lit to be a string value that we pass directly into _GrumpRun and then compare the output with `repr(ast.literal_eval(lit))`.

I think the effect is equivalent for the other kinds of literals that we test, but let me know if you think that's not the case.

compiler/expr_visitor_test.py
```diff
@@ -132,6 +134,8 @@ def foo(a, b=2):
   testDictEmpty = _MakeLiteralTest({})
   testDictNonEmpty = _MakeLiteralTest({'foo': 42, 'bar': 43})
 
+  testSetNoneEmpty = _MakeLiteralTest({'foo', 'bar'}, "set(['foo', 'bar'])")
```
trotterdylan: testSetNonEmpty

compiler/expr_visitor.py
```diff
@@ -178,6 +178,17 @@ def visit_Dict(self, node):
       self.writer.write('{} = {}.ToObject()'.format(result.name, d.expr))
     return result
 
+  def visit_Set(self, node):
+    with self.block.alloc_temp("*πg.Set") as s:
```
trotterdylan: Please use single quotes in this file

compiler/util.py
```diff
@@ -101,6 +101,12 @@ def write_checked_call2(self, result, call, *args, **kwargs):
         \tcontinue
         }"""), result=result.name, call=call.format(*args, **kwargs))
 
+  def write_checked_call2_as_temp(self, call, *args, **kwargs):
```
aisk: Hi, I added the `blank_expr` as an instance of  `GeneratedLiteral`, but looks like `write_checked_call2` need a instance with a name property, so I added the name property to `GeneratedLiteral`, I don't know whether it's suitable or not.

compiler/expr_visitor_test.py
aisk: Seems current test case have the same problem, for example, `_MakeLiteralTest(1E6)`, the CPython parser will parse it to `100000.0` before we run it. I added a new function `_MakeLiteralStringTest`, should we change all the current tests to it ?

aisk:
@trotterdylan Hi I modified the PR, and have some comments above, can you give a review again? thanks!

compiler/util.py
trotterdylan: I see. I would actually create a new GeneratedExpr class:

```
class BlankVar(object):
  def __init__(self):
    self.name = '_'
    self.expr = '_'

blank_var = BlankVar()
```

compiler/expr_visitor_test.py
```diff
@@ -132,6 +140,8 @@ def foo(a, b=2):
   testDictEmpty = _MakeLiteralTest({})
   testDictNonEmpty = _MakeLiteralTest({'foo': 42, 'bar': 43})
 
+  testSetNoneEmpty = _MakeLiteralStringTest({'foo', 'bar'})
```
trotterdylan: I think you want to pass a string value into _MakeLiteralStringTest() right? otherwise this value will be converted to the string "set(['foo', 'bar'])" which will be passed into _GrumpRun().

Otherwise, I think this approach is good, and yes, I think all the literal tests should work this way. So maybe just change the way _MakeLiteralTest works.

aisk:
Moded all old tests to new approach, but looks like current float type's repr implement is different from cpython's.

trotterdylan:
You could mark the test with [expectedFailure](https://docs.python.org/2/library/unittest.html#unittest.expectedFailure) for now and we can file an issue to track fixing it. The goal for Grumpy is to be highly compatible with CPython so this is definitely a bug.

I think your approach here is reasonable because the important thing to test is the compiler, not Grumpy's repr implementation. That said, since the goal is repr compatibility as well, it wouldn't hurt if that was implicitly tested as well. So perhaps add a TODO to the assertEqual that describes why it's done this way and how things would be different in an ideal world.

aisk:
Just marked these tests with expectedFailure now, and added an TODO on them. I think current test approach is most straightforward before the repr bug was fixed.

aisk:
@trotterdylan Hi i think this PR and #232 can be merged now.

PR #209, opened 2017-01-25
Add datetime module and unittest

S-YOU:
- datetime.py is from pypy, which seems to be from CPython3.
- No locale support
- Timezone support have to confirm.

```
% make run
import time
import datetime
timedelta, date = datetime.timedelta, datetime.date
d = timedelta(microseconds=-1)
print (d.days, d.seconds, d.microseconds)

today = date.today()
print today
print today == date.fromtimestamp(time.time()

^D
(-1, 86399, 999999)
2017-02-16
True
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,7 @@ STDLIB_TESTS := \
   test/test_string \
   test/test_md5 \
   test/test_bisect \
+  test/test_datetime \
   threading_test \
   time_test \
   types_test \
--- a/third_party/pypy/datetime.py
+++ b/third_party/pypy/datetime.py
@@ -1,0 +1,2102 @@
+"""Concrete date/time and related types -- prototype implemented in Python.
+
+See http://www.zope.org/Members/fdrake/DateTimeWiki/FrontPage
+
+See also http://dir.yahoo.com/Reference/calendars/
+
+For a primer on DST, including many current DST rules, see
+http://webexhibits.org/daylightsaving/
+
+For more about DST than you ever wanted to know, see
+ftp://elsie.nci.nih.gov/pub/
+
+Sources for time zone and DST data: http://www.twinsun.com/tz/tz-link.htm
+
+This was originally copied from the sandbox of the CPython CVS repository.
+Thanks to Tim Peters for suggesting using it.
+"""
+
+# from __future__ import division
+import time as _time
+import math as _math
+# import struct as _struct
+import _struct
+
+def divmod(x, y):
+  x, y = int(x), int(y)
+  return x / y, x % y
+
+_SENTINEL = object()
+
+def _cmp(x, y):
+    return 0 if x == y else 1 if x > y else -1
+
+def _round(x):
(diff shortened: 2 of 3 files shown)


kylez420:
James

third_party/misc/brython_LICENCE.txt
```diff
@@ -0,0 +1,10 @@
+Copyright (c) 2012, Pierre Quentel pierre.quentel@gmail.com
```
trotterdylan: How about third_party/brython?

third_party/misc/brython_time.py
```diff
@@ -0,0 +1,406 @@
+# from browser import window
```
trotterdylan: I'm a little worried about using this implementation since it depends on Brython-specific functionality. Seems like more effort to work around potential incompatibilities here than to just write something from scratch. I might be underestimating the effort involved though.

third_party/stdlib/dummy_thread.py
```diff
@@ -0,0 +1,145 @@
+"""Drop-in replacement for the thread module.
```
trotterdylan: This is good to have since it seems to be used by some other modules, but I think it should be pretty straightforward to implement a thread.py. See threading.py for some examples.

third_party/misc/brython_time.py
S-YOU: Write something from scratch is definitely possible, since strftime implementation is simple, I was just not sure using third party vs writing from scratch in terms of maintainability, but looks like that's not the case for this one.

third_party/stdlib/dummy_thread.py
S-YOU: thread.py is already in grumy, I was just not sure what's that lock function does.

trotterdylan: Lock is similar to Mutex in Go. In the short term I think something like this would work:

```
class Lock(object):
  def __init__(self):
    self._mutex = Mutex.new()

  def acquire(self):
    self._mutex.Lock()

  def release(self):
    self._mutex.Unlock()

  def __enter__(self):
    self.acquire()

  def __exit__(self, *args):
    self.release()
```

Long term there are some behavior differences to work around since acquire() can be non-blocking, but Go mutex does not support that mode.

S-YOU: Thanks for the explanation and code.

S-YOU:
I wrote strftime and struct_time implementation, please take a look.
If the approach is ok I will continue to work on adding tests from stdlib.

trotterdylan:
This is looking good. Would you mind doing the thread and time changes in a separate PR? I think it'd be easier to review that way. Thanks!

S-YOU:
Sure

S-YOU:
Done at #218, and #219

S-YOU:
I have updated this PR to datetime module only, and added unittest from stdlib.

PR #212, opened 2017-01-25
Count methods

meadori:
This PR implement count methods for `str` and `tuple`.  The `str` one is slight different b/c `str`s need to account for non-overlapping substrings.  Luckly, Go's `strings.Count` does the right thing.


runtime/str_test.go
```diff
@@ -137,6 +137,19 @@ func TestStrContains(t *testing.T) {
 	}
 }
 
+func TestStringCount(t *testing.T) {
+	cases := []invokeTestCase{
+		{args: wrapArgs("", "a"), want: NewInt(0).ToObject()},
+		{args: wrapArgs("abcdeffdeabcb", "b"), want: NewInt(3).ToObject()},
+		{args: wrapArgs(""), wantExc: mustCreateException(TypeErrorType, "count() takes exactly one argument (1 given)")},
```
trotterdylan: Also worth testing wrapArgs("foo", "")

runtime/str_test.go
```diff
@@ -137,6 +137,19 @@ func TestStrContains(t *testing.T) {
 	}
 }
 
+func TestStringCount(t *testing.T) {
```
trotterdylan: Any reason not to put this in TestStringMethods?

runtime/tuple.go
```diff
@@ -75,6 +75,14 @@ func tupleContains(f *Frame, t, v *Object) (*Object, *BaseException) {
 	return seqContains(f, t, v)
 }
 
+func tupleCount(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	argc := len(args)
+	if argc != 2 {
+		return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("count() takes exactly one argument (%d given)", argc))
```
trotterdylan: I think you should check that args[0] is a tuple so probably worth using checkMethodArgs()

runtime/tuple_test.go
```diff
@@ -102,6 +102,19 @@ func TestTupleContains(t *testing.T) {
 	}
 }
 
+func TestTupleCount(t *testing.T) {
+	cases := []invokeTestCase{
+		{args: wrapArgs(NewTuple(), NewInt(1)), want: NewInt(0).ToObject()},
+		{args: wrapArgs(NewTuple(None, None, None), None), want: NewInt(3).ToObject()},
+		{args: wrapArgs(NewTuple()), wantExc: mustCreateException(TypeErrorType, "count() takes exactly one argument (1 given)")},
```
trotterdylan: Also a test case where the first arg is not a tuple

runtime/str_test.go
meadori: Just an oversight on my part :smile:

runtime/str_test.go
meadori: ACK.  Will fix.

meadori: I have a similar test in `testing/str_test.py`.

runtime/tuple.go
meadori: I did the hand-coded error check so that I could match the CPython error message.  Plus, isn't the type error already caught?:

```
$ cat foo.py 
print tuple.count("ab", 2)
$ ./build/bin/grumpc foo.py > foo.go
$ go run foo.go                     
TypeError: unbound method count() must be called with tuple instance as first argument (got str instance instead)
exit status 1
```

Also, I did the exact same style check in `strCount` and `listCount`.  I would prefer to do it the same way everywhere.

trotterdylan: Agree about doing the same everywhere. And you're right that the common case will be caught by the method type check you mention. The trickier case is if somebody gets the underlying function via instancemethod's im_func attribute:

```
>>> class A(object):
...  def foo(self):
...   pass
...
>>> A.foo
<unbound method A.foo>
>>> A.foo.im_func
<function foo at 0x10407f140>
```

Note that its not possible to get the underlying function for tuple.count in CPython because builtin methods are not instancemethod objects there:

```
>>> tuple.count
<method 'count' of 'tuple' objects>
>>> tuple.count.im_func
AttributeError: 'method_descriptor' object has no attribute 'im_func'
```

However, in Grumpy (at least for the time being), builtin methods *are* instancemethods. So it's possible for somebody to get the underlying function for tuple.count and call that directly. So the redundant check is actually important here.

runtime/tuple_test.go
trotterdylan: This won't actually exercise the code that I'm concerned about that we're discussing above, so adding this test is probably not very useful.

meadori:
The last commit I pushed should have addressed any issues.

Issue #210, opened 2017-01-25
"make install" fails on Mac OS X


cclauss:
Mac OS X install command does not accept -D (capital D)

$ install
usage: install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
               [-o owner] file1 file2
       install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
               [-o owner] file1 ... fileN directory
       install -d [-v] [-g group] [-m mode] [-o owner] directory ...

trotterdylan:
From man install: -D create all leading components of DEST except the last, then copy SOURCE to DEST

Looks like we could add an extra step to the recipe to just do `mkdir -p` on dirname of the target.

cclauss:
Closed by https://github.com/google/grumpy/pull/213

PR #213, opened 2017-01-25
Changed install section of Makefile to work on Mac OS X

cclauss:
Usage: `make DESTDIR=<dest_dir> install`

See https://github.com/google/grumpy/issues/210

* Mac OSX `install` does not accept the `-D` option
* Mac OSX `cp` does not accept the `--no-preserve=ownership` option

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,12 +289,12 @@ $(BENCHMARK_BINS): build/benchmarks/%_benchmark: build/benchmarks/%.go $(RUNTIME
 # ------------------------------------------------------------------------------
 
 install: $(RUNNER_BIN) $(COMPILER) $(RUNTIME) $(STDLIB)
-        mkdir -p "$(DESTDIR)/usr/bin"
-        # Binary executables
-        install -m755 build/bin/grumpc "$(DESTDIR)/usr/bin/grumpc"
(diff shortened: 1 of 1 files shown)


cclauss:
I do not know make well.  Perhaps @xyproto could tell us if the changes that I made in the install_osx section break anything that he was trying to achieve in the normal install section.  If not, then we could try using the new code on all platforms.

trotterdylan:
I think what you have will work fine on Linux. The only thing I don't quite understand the importance of is `--no-preserve=ownership`.

xyproto:
Removing `--no-preserve=ownership` is probably fine. It's just a precaution in case files that are to be installed have unusual file ownership.

trotterdylan:
Thanks for the info Alexander. That makes sense about using install -d vs. mkdir -p.

@cclauss looks like there's a syntax error in the Makefile currently according to Travis: `Makefile:292: *** missing separator (did you mean TAB instead of 8 spaces?).`

cclauss:
I fixed the spaces --> tab problem and removed the `mkdir -p` and replaced with `install -d` (lowercase d ;-) ) as recommended by @xyproto

Issue #211, opened 2017-01-25
Cannot build on macOS 10.12.2


miguelammatos:
$ git show --summary
commit e91485cb544304d67e4c240f8ac77aa72072248a

$ make clean

$ make
build/src/grumpy/builtin_types.go:780: z.Text undefined (type big.Int has no field or method Text)
build/src/grumpy/long.go:144: x.Text undefined (type *big.Int has no field or method Text)
build/src/grumpy/long.go:298: toLongUnsafe(o).value.Text undefined (type big.Int has no field or method Text)
build/src/grumpy/long.go:306: toLongUnsafe(o).value.Text undefined (type big.Int has no field or method Text)
build/src/grumpy/str.go:855: toLongUnsafe(o).Value().Text undefined (type *big.Int has no field or method Text)
make: *** [build/pkg/darwin_amd64/grumpy.a] Error 1

$ /usr/bin/env python -V
Python 2.7.3

trotterdylan:
What version of go are you running: go version

miguelammatos:
$go version
go version go1.5.1 darwin/amd64

trotterdylan:
I think the minimum required version is either 1.6 or 1.7.

miguelammatos:
Thanks, I'll update and recheck.

miguelammatos:
It works with version 1.7.4.
It would be nice to have this in the requirements, I've submitted a PR that checks this on the Makefile similarly to what done for python.

PR #214, opened 2017-01-25
Check for minimum Go version, see #211.

miguelammatos:
Add validation of minimum Go version using only make language for portability.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,15 +32,18 @@ ifeq ($(PYTHON),)
 endif
 PYTHON_BIN := $(shell which $(PYTHON))
 PYTHON_VER := $(word 2,$(shell $(PYTHON) -V 2>&1))
-GO_MIN_VER := 1.5
-GO_VER := $(subst go,, $(word 3,$(shell go version 2>&1)) )
+GO_REQ_MAJ := 1
+GO_REQ_MIN := 6
+GO_MAJ_MIN := $(subst go,, $(word 3,$(shell go version 2>&1)) )
+GO_MAJ := $(word 1,$(subst ., ,$(GO_MAJ_MIN) ))
+GO_MIN := $(word 2,$(subst ., ,$(GO_MAJ_MIN) ))
 
 ifeq ($(filter 2.7.%,$(PYTHON_VER)),)
   $(error unsupported Python version $(PYTHON_VER), Grumpy only supports 2.7.x. To use a different python binary such as python2, run: 'make PYTHON=python2 ...')
 endif
 
-ifneq "$(GO_MIN_VER)" "$(word 1, $(sort $(GO_MIN_VER) $(GO_VER)))"
-  $(error unsupported Go version $(GO_VER), Grumpy requires at least $(GO_MIN_VER).x. Please update Go.)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,11 +32,17 @@ ifeq ($(PYTHON),)
 endif
 PYTHON_BIN := $(shell which $(PYTHON))
 PYTHON_VER := $(word 2,$(shell $(PYTHON) -V 2>&1))
+GO_MIN_VER := 1.5
```
trotterdylan: Were you able to confirm that it works with 1.6? Be conservative and require 1.7 if not.

cclauss: I thought the minimum was v1.6, not v1.5.

Makefile
```diff
@@ -32,11 +32,17 @@ ifeq ($(PYTHON),)
...
+GO_VER := $(subst go,, $(word 3,$(shell go version 2>&1)) )
 
 ifeq ($(filter 2.7.%,$(PYTHON_VER)),)
   $(error unsupported Python version $(PYTHON_VER), Grumpy only supports 2.7.x. To use a different python binary such as python2, run: 'make PYTHON=python2 ...')
 endif
 
+ifneq "$(GO_MIN_VER)" "$(word 1, $(sort $(GO_MIN_VER) $(GO_VER)))"
```
trotterdylan: This is clever! It took me a minute to figure it out :)

I think it'd be better if the comparison was more explicit:

```
GO_VER := $(shell go version | sed 's/.*go\([0-9][0-9]*\).\([0-9][0-9]*\).*/\1 \2/')
GO_VER_MAJOR := $(word 1,$(GO_VER))
GO_VER_MINOR := $(word 2,$(GO_VER))
```

And then somehow use `test $(GO_VER_MAJOR) -lt 1 || test $(GO_VER_MINOR) -lt 6` to do the comparison.

Makefile
miguelammatos: Cannot test with 1.6 right now, I'll do in a couple of hours.

miguelammatos: Yes, my mistake, I'll update it.

Makefile
miguelammatos: If we can assume that sed is going to be present then yes, that approach is more readable.
Otherwise the proposed approach only depends on make.
To check for major, minor we can do something like.

`
GO_VER := $(subst go,, $(word 3,$(shell go version 2>&1)) )
GO_MAJ_VER := $(word 1,$(subst ., ,$(GO_VER) ))
GO_MIN_VER := $(word 2,$(subst ., ,$(GO_VER) ))`

and then compare accordingly.

trotterdylan: That sounds good to me. Thanks for iterating on this.

Makefile
miguelammatos: Works in 1.6 (tested under ubuntu).
Updated version accordingly.

Makefile
miguelammatos: New push with the proposed changes.

miguelammatos:
Thanks, added the commit e-mail address to cla, it shouldn't take long.

PR #216, opened 2017-01-26
Replace use of builtin ast module with pythonparser.

trotterdylan:
[pythonparser](https://github.com/m-labs/pythonparser) is a high quality Python parser written in Python that produces an AST very similar to the one produced by the standard ast module. There are three very appealing qualities to switching Grumpy to pythonparser:

1. It's pure Python so there's a reasonable chance of getting it to build under Grumpy in the near future, whereas the ast module, being written in C, is unlikely to be work anytime soon.

2. The AST produced by pythonparser contains detailed span location data. The standard ast module provides line numbers only and not for all nodes. pythonparser provides column data, sub-node locations, locations for comments and more.

3. Experimenting with custom syntax (e.g. to properly support importing Go packages with arbitrary names, like `from __go__ import "github.com/labstack/echo"`) will become much simpler.

I suspect pythonparser is quite a bit slower than the ast module but it doesn't seem to matter since Go compilation dominates. Timings for `make test` were 9m35.968s under the ast module and 10m7.250s under pythonparser.

There are a few minor outstanding issues with the library [I've filed](https://github.com/m-labs/pythonparser/issues/created_by/trotterdylan). These seem relatively straightforward to fix and the maintainer is open to PRs. Overall I'm very satisfied with the quality and completeness of the library.


nairb774:
Ooo, nice. This is great. It would be so cool to be able to `go get` the compiler and runtime, and just start transforming/compiling code.

S-YOU:
That's Awesome, to get it working. The issue relating decorator on that parser can work-around?

alanjds:
Tiny suggestion: 

I skimmed of the diff and saw lots of ``node.lineno`` replaced by ``node.loc.line()``.

Would it be better to land a patch on pythonparser with a property, raising the compatibility over the official module and lowering the size of patches here?

trotterdylan:
Yeah that's something I considered. I haven't brought it up with the maintainers but I might see what a patch looks like a file a PR if it's simple. Thanks for bringing this up.

alanjds:
I started a [stdlib-compat-ast branch](https://github.com/m-labs/pythonparser/compare/master...alanjds:stdlib-compat-ast?expand=1) on pythonparser. For now it only imports the ``test_ast.py`` from CPython repo (2.7 tag) and fixes ``.parse()`` on code not ended by ``\n``.

Lots of tests are broken, but will at least try to fix the compat for the things Grumpy uses. I have not much free time this days but will update you tomorrow, ok?

trotterdylan:
That's fantastic. I like the approach. Thanks for getting that started.

alanjds:
I had some progress on the branch. Please take a new look on the diff link.

Lots of tests still breaks but the only harmful one is ``test_slice`` I guess. Here it go: 
```
(...)
======================================================================
FAIL: test_slice (pythonparser.test.test_ast.AST_Tests)
  File "/Users/alanjds/src/git/pythonparser/pythonparser/test/test_ast.py", line 230, in test_slice
    self.assertIsInstance(slc.step, ast.Name)
AssertionError: None is not an instance of <class 'pythonparser.ast.Name'>

Ran 138 tests in 7.233s

FAILED (failures=3, errors=11)
```

trotterdylan:
3 failures and 11 errors isn't bad at all!

I would recommend breaking out the trailing `\n` fix and the lineno property addition and sending them as two PRs. Getting all the ast tests working may take some time, but those two changes would make my PR much more manageable.

alanjds:
Well... the are about 30 tests on test_ast, and about half is broken. Two "suits" are in it: AST_Tests and ASTHelpers_Test. I guess that ASTHelpers_Test is not needed to be fixed because it test internals and specifics, but AST_Tests do test most of the direct funcionality.

Will ask there (pythonparser) what to do and mention this issue there. (but maybe not today)

alanjds:
Made 3 PRs on pythonparser, one for each compat feature fixed. And one for the new ``compat_ast`` module plus tests (https://github.com/m-labs/pythonparser/issues/16).

Now is time to wait, I guess.

trotterdylan:
I saw those PRs fly by. Thanks for sending them!

whitequark:
FYI I've fixed all outstanding issues except https://github.com/m-labs/pythonparser/issues/6. It seems like #6 will be handled best with a PR with the lexer patch that you already have.

PR #218, opened 2017-01-26
Add LockType, and allocate_lock to thread.py

S-YOU:


meadori:
I started [implementing something similar](https://github.com/meadori/grumpy/tree/threading), but ended up deciding that it would have to be implemented in Go instead of pure Python b/c of non-blocking lock support (which Go Mutex don't natively support).  There is a very gross hack in my WIP commit that implements non-blocking locks.

I think the way to go with the `threading` library is to create the basic `Lock` type in Go, then layer everything else on top of that in Python.  This is essentially what CPython does and if we do it right, then we should be able to copy most (all?) of the `threading` module from CPython.

S-YOU:
My commit only contain codes given by @trotterdylan on another thread, so I will leave him to decide.

trotterdylan:
I definitely agree the implementation isn't ideal. The implementation of RecursiveLock will have to be in Go one way or the other, so it probably makes sense to implement both kinds of locks there. In the short term, though, I think it's reasonable to not support the blocking arg. The code should fail hard when somebody tries to pass blocking=True and then we'll know when we need to work on a more sophisticated implementation.

PR #219, opened 2017-01-26
Add struct_time, gmtime, localtime, mktime, strftime to lib/time

S-YOU:


lib/time.py
```diff
@@ -23,3 +55,37 @@ def sleep(secs):
...
+    'Z':  'MST', #'z': '-0700',
+    '%': '%',
+    # TODO: Implement c, x, X, j, U, w, W
+}
+
+def strftime(layout, tt=None): #pylint: disable=missing-docstring
+  ret = ''
+  t = (Unix(int(mktime(tt)), 0) if tt else Now()).Local()
+  prev, n = 0, layout.find('%')
+  while 0 <= n <= len(layout) - 2:
```
trotterdylan: I like the overall approach of this function. I just have a few comments about specific details that are all kind of related:

* The way the loop condition works right now, there are two reasons why it might exit (% was found at the end of the string or % was not found.) I think it'd be easier to understand if there was one condition: `n != -1`
* It's cheaper to append the pieces of the result to a list and join them at the end.
* It seems like strftime discards the % for escapes that it doesn't recognize so I think you can remove the TODO:

```
>>> time.strftime('%3')
'3'
```

* The logic is a little simpler if you use _strftime_map.get() instead of checking `next_val in _strftime_map`.

I've written an example implementation (just to prove to myself that it's simpler, no need to copy it verbatim) that demonstrates these ideas:

```
ret = []
prev, n = 0, layout.find('%', 0, -1)
while n != -1:
  ret.append(layout[prev:n])
  c = _strftime_map.get(layout[n+1])
  if c:
    ret.append(t.Format(c))
  else:
    ret.append(c)
  prev, n = n, layout.find('%', n + 2, -1)
ret.append(layout[prev:])
return ''.join(ret)
```

lib/time.py
```diff
@@ -23,3 +55,37 @@ def sleep(secs):
...
+
+
+_strftime_map = {
+    'Y': '2006', 'y': '06', 'b': 'Jan', 'B': 'January', 'm': '01', 'd': '02',
+    'a': 'Mon', 'A': 'Monday',
+    'H': '15', 'I': '03', 'M': '04', 'S': '05', 'L': '.000',
+    'p': 'PM',
+    'Z':  'MST', #'z': '-0700',
+    '%': '%',
+    # TODO: Implement c, x, X, j, U, w, W
```
trotterdylan: Maybe these should raise NotImplemented

lib/time.py
```diff
@@ -23,3 +55,37 @@ def sleep(secs):
 
 def time():
   return float(Now().UnixNano()) / Second
+
+
+_strftime_map = {
```
trotterdylan: I would call this _strftime_directive_map or something.

lib/time.py
```diff
@@ -23,3 +55,37 @@ def sleep(secs):
...
+    'Y': '2006', 'y': '06', 'b': 'Jan', 'B': 'January', 'm': '01', 'd': '02',
+    'a': 'Mon', 'A': 'Monday',
+    'H': '15', 'I': '03', 'M': '04', 'S': '05', 'L': '.000',
+    'p': 'PM',
+    'Z':  'MST', #'z': '-0700',
+    '%': '%',
+    # TODO: Implement c, x, X, j, U, w, W
+}
+
+def strftime(layout, tt=None): #pylint: disable=missing-docstring
```
trotterdylan: In CPython, the time module calls the first parameter `format`.

S-YOU: I will change that, I didn't use it because format seems to be python's builtins function.

lib/time.py
S-YOU: Regarding %3, I got like following on CPython, 
>>> time.strftime('%3')
' %3'

S-YOU: dict.get return TypeError in grumpy
```
c = _strftime_directive_map.get(format[n+1])
TypeError: 'in <string>' requires string as left operand, not NoneType
```
so I think I have to put next_char into variable and use `.get(next_char, next_char)`

S-YOU:
Updated based on some reviews, but format `%3` has returning strange output, still looking into it.

S-YOU:
Updated based on some reviews, except %3 discrepancy.

Your result: '3'
My CPython: ' %3'
What I did: '%3'

S-YOU:
Raising NotImplementedError on %number case, Please take a look again.

lib/time.py
trotterdylan: That sounds fine. I think the TypeError is a bug of some kind though.

S-YOU: Looks like it is from next line, strContains, looks like we have one number off in lineno.

S-YOU: The code I submitted last not using in operator on that place, so removed that ''.

lib/time.py
```diff
@@ -23,3 +55,39 @@ def sleep(secs):
...
+}
+
+def strftime(format, tt=None): #pylint: disable=missing-docstring
+  t = (Unix(int(mktime(tt)), 0) if tt else Now()).Local()
+  ret = []
+  prev, n = 0, format.find('%', 0, -1)
+  while n != -1:
+    ret.append(format[prev:n])
+    next_ch = format[n+1]
+    if next_ch in 'cjUwWxX0123456789':
```
trotterdylan: I think 0123... don't need to be specialized. Anything that isn't a ftime directive should just be expanded to its normal value:

```
>>> time.strftime('%3%4%o')
'34o'
```

trotterdylan: I would put the cjU... chars in the directive map with a special value (e.g. NotImplemented) and check specially for that. That way the unimplemented directives will be grouped together in the map with the rest and it'll be clear to reader what works and what's missing.

Also, if the key is absent from the map, it means it should be passed through directly (e.g. %o)

S-YOU: In my tests on CPython, they work like padding, I didnt get 3 for %3 but ' %3' instead.

lib/time.py
```diff
@@ -23,3 +55,39 @@ def sleep(secs):
...
+_strftime_directive_map = {
+    'Y': '2006', 'y': '06', 'b': 'Jan', 'B': 'January', 'm': '01', 'd': '02',
+    'a': 'Mon', 'A': 'Monday',
+    'H': '15', 'I': '03', 'M': '04', 'S': '05', 'L': '.000',
+    'p': 'PM',
+    'Z':  'MST', 'z': '-0700',
+    '%': '%',
+}
+
+def strftime(format, tt=None): #pylint: disable=missing-docstring
```
S-YOU: Missed again, thank you

lib/time.py
S-YOU: http://ideone.com/wFMiXv (python2), http://ideone.com/PZKsHs (python3)
@trotterdylan, what version or implementation of python do you have?

trotterdylan: My system Python is 2.7.10:

```
$ python
Python 2.7.10 (default, Oct 23 2015, 19:19:21)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> print [time.strftime('%3')]
['3']
print [time.strftime('%3%4%o')]>>> print [time.strftime('%3%4%o')]
['34o']
^D
$ python3
Python 3.6.0 (default, Jan 10 2017, 20:25:59)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> print([time.strftime('%3')])
['3']
print([time.strftime('%3%4%o')])
>>> print([time.strftime('%3%4%o')])
['34o']
```

ideaone seems to use some weird implementation.

S-YOU: @trotterdylan , I could reproduce that in my OSX, but linux is different.
Could you try on linux too?

S-YOU: May be, related open discussion python - https://bugs.python.org/issue9811

trotterdylan: Good find! I get the same results on my Linux box.

It seems like it's a limitation of the spec: `If a conversion specification does not correspond to any of  the  above,  the behavior is undefined.`

Let's aim to match the Linux behavior.

S-YOU: Thanks for confirmation

S-YOU: To match Linux behavior, I think I still need 0-9 in NotImpl list, should I continue write 17 NotImpl in the directive map?

trotterdylan: I think I'm missing something. 0-9 should just be output verbatim, right? So just leave them out of the directive map and pass through any value that's not present in the directive map. There are other values that need to be output verbatim as well, e.g.:

```
>>> time.strftime('%f')
'%f'
```

S-YOU: This differ between OSX and Linux too.
Linux treat 0-9 as zero padding as far as I tried. Please try `%5H` on linux.

```
% python
Python 2.7.9 (default, Jun 29 2016, 13:08:31) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> time.strftime('%5H')
'00017'
>>> 
```

My above implementation do pass through value that not present in the map like %f %o.

trotterdylan: Argh. Thanks for clarifying. It's annoying how inconsistent it all is:

- According to [man strftime](http://man7.org/linux/man-pages/man3/strftime.3.html), Linux supports padding and 0 flag for zero-padding
- [The UNIX spec](http://pubs.opengroup.org/onlinepubs/007908799/xsh/strftime.html) does not mention any kind of padding
- [GNU strftime](http://search.cpan.org/~dexter/POSIX-strftime-GNU-0.02/lib/POSIX/strftime/GNU.pm) supports 0 padding but not arbitrary length padding (same as for [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=strftime&sektion=3) which explains the way OS X behaves)
- And of course [time.strftime()](https://docs.python.org/2/library/time.html#time.strftime) defers to the platform specific behavior

Thinking about this some more, I'm inclined to make this as simple as possible while still adhering to the Python spec. I don't see any value in supporting any extensions specific to one particular platform.

As far as whether '%f' -> 'f' or '%f' -> '%f', I don't have a strong opinion but I'd probably lean toward Linux since it's so widely used for server development.

S-YOU:
Thanks for digging details into this and reviews iterations.
Rebased but didn't squash commits for easy diff.

I have changed to use special value `NotImplementedError` for `NotImplementedError` for not supported values, please let me know if you want something else instead of `NotImplementedError` as value.

lib/time.py
```diff
@@ -23,3 +55,48 @@ def sleep(secs):
...
+  t = (Unix(int(mktime(tt)), 0) if tt else Now()).Local()
+  ret = []
+  prev, n = 0, format.find('%', 0, -1)
+  while n != -1:
+    ret.append(format[prev:n])
+    next_ch = format[n+1]
+    c = _strftime_directive_map.get(next_ch)
+    if c is NotImplementedError:
+      raise NotImplementedError('Code: %' + next_ch + ' not yet supported')
+    elif c:
```
trotterdylan: This can be an `if` rather than `elif` since the previous if clause exited the function.

lib/time.py
```diff
@@ -14,7 +14,39 @@
...
+            "tm_yday=%s, tm_isdst=%s)") % self
+
+  def __str__(self):
+    return repr(self)
+
+
+def gmtime(seconds=None):
+  t = (Unix(seconds, 0) if seconds else Now()).UTC()
+  return struct_time((t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(),
+                      t.Second(), (t.Weekday()+6)%7, t.YearDay(), 0))
```
trotterdylan: Put spaces around operators:

```
t.Second(), (t.Weekday() + 6) % 7, t.YearDay(), 0))
```

lib/time.py
```diff
@@ -14,7 +14,39 @@
...
+
+
+def localtime(seconds=None):
+  t = (Unix(seconds, 0) if seconds else Now()).Local()
+  return struct_time((t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(),
+                      t.Second(), (t.Weekday()+6)%7, t.YearDay(), 0))
```
trotterdylan: Spaces around operators

lib/time.py
```diff
@@ -23,3 +55,48 @@ def sleep(secs):
 
 def time():
   return float(Now().UnixNano()) / Second
+
+
+_strftime_directive_map = {
+    'Y': '2006', 'y': '06', 'b': 'Jan', 'B': 'January', 'm': '01', 'd': '02',
```
trotterdylan: Would you mind reformatting this dict to be one directive per line? Although it takes up more space, I think it'll be easier to read. Also, I would put the keys in alphabetical order.

lib/time.py
```diff
@@ -23,3 +55,48 @@ def sleep(secs):
...
+    '9': NotImplementedError,
+}
+
+def strftime(format, tt=None):  # pylint: disable=missing-docstring,redefined-builtin
+  t = (Unix(int(mktime(tt)), 0) if tt else Now()).Local()
+  ret = []
+  prev, n = 0, format.find('%', 0, -1)
+  while n != -1:
+    ret.append(format[prev:n])
+    next_ch = format[n+1]
```
trotterdylan: Spaces around operators, even inside subscript (this is different than the way subscripts are formatted in Go).

lib/time.py
```diff
@@ -14,7 +14,39 @@
...
 
-from __go__.time import Now, Second, Sleep  # pylint: disable=g-multiple-import
+from __go__.time import Now, Second, Sleep, Unix, Date, UTC # pylint: disable=g-multiple-import
+
+
+class struct_time(tuple):  #pylint: disable=invalid-name,missing-docstring
+
+  def __init__(self, args):
+    super(struct_time, self).__init__(tuple, args)
+    self.tm_year, self.tm_mon, self.tm_mday, self.tm_hour, self.tm_min, \
```
trotterdylan: Do assignments one per line please.

lib/time.py
S-YOU: Sure

lib/time.py
S-YOU: I see, thanks for the explanations.

lib/time.py
S-YOU: Yeah, I think I confused here.

S-YOU:
Review updates done. I would like to add stdlib unittest for this, but those are heavily rely on string formatting with %02d or similar, so I would like to wait until #221 is ok.

lib/time.py
```diff
@@ -23,3 +62,67 @@ def sleep(secs):
...
+    'w': NotImplementedError,
+    'X': NotImplementedError,
+    'x': NotImplementedError,
+    'y': '06',
+    'Y': '2006',
+    'Z': 'MST',
+    'z': '-0700',
+}
+
+def strftime(format, tt=None):  # pylint: disable=missing-docstring,redefined-builtin
```
trotterdylan: Two blank lines before module-level functions and classes please. Probably makes sense to move the directive map to the top of the file so that the module functions are all grouped together.

lib/time.py
```diff
@@ -23,3 +62,67 @@ def sleep(secs):
 
 def time():
   return float(Now().UnixNano()) / Second
+
+
+_strftime_directive_map = {
+    '%': '%',
+    '0': NotImplementedError,
```
trotterdylan: Since we're not implementing padding, we can remove these, right?

S-YOU: That will be different behavior with cpython on linux, if you are ok, I am ok

lib/time.py
S-YOU: Sure

lib/time.py
S-YOU: Linux output following by the way, 
```
>>> time.strftime('%10')
'       %10'
```
if I remove from that list, output of `'%10'` will be 
```
'%10'
```

trotterdylan: Right I think we should leave out platform specific extensions like padding. From my previous comment:

"Thinking about this some more, I'm inclined to make this as simple as possible while still adhering to the Python spec. I don't see any value in supporting any extensions specific to one particular platform."

lib/time.py
```diff
@@ -23,3 +100,31 @@ def sleep(secs):
...
+
+def strftime(format, tt=None):  # pylint: disable=missing-docstring,redefined-builtin
+  t = (Unix(int(mktime(tt)), 0) if tt else Now()).Local()
+  ret = []
+  prev, n = 0, format.find('%', 0, -1)
+  while n != -1:
+    ret.append(format[prev:n])
+    next_ch = format[n + 1]
+    c = _strftime_directive_map.get(next_ch)
+    if c is NotImplementedError:
```
trotterdylan: You could use the builtin NotImplemented if you wanted to save a few characters.

S-YOU: Sure

lib/time.py
S-YOU: Got it

S-YOU:
Updated latest reviews.

PR #220, opened 2017-01-26
Create Standard Library checklist.md

cclauss:
```python
#!/usr/bin/env python3
import bs4, requests
url = 'https://docs.python.org/2.7/py-modindex.html'
soup = bs4.BeautifulSoup(requests.get(url).content, 'html.parser')
for module in soup.find_all('code', class_='xref'):
    if '.' not in module.string:
        print('- [ ] ' + module.string.replace('__', '\_\_', 1))
```


meadori:
I think this is useful to add.  Maybe add an `owner` column too?  This way folks could sign up for various libraries.  I would hate for multiple people to spend lots of time porting the same library.  I would be curious what @trotterdylan thinks about how to organize this.

trotterdylan:
Thanks for doing this! This list will be useful. We have a (very limited) list in https://github.com/google/grumpy/wiki/Missing-features. Can we consolidate this information into that wiki page instead?

cclauss:
Putting the checklist into to wiki is fine.

trotterdylan:
Are you able and willing to propose a PR for the wiki change?

S-YOU:
I don't see edit button or find a way to send PR by the way, may be I am missing something.

http://stackoverflow.com/questions/10642928/how-to-pull-request-a-wiki-page-on-github
http://roman-ivanov.blogspot.jp/2013/11/how-to-merge-github-wiki-changes-from.html

trotterdylan:
Okay, I've updated the [wiki page](https://github.com/google/grumpy/wiki/Missing-features#standard-libraries) to add the list that cclaus provided. I've removed some of the platform specific deprecated libraries as well. Thanks all!

cclauss:
Cool... Thx

cclauss:
What should we do about cPickle, cProfile, cStringIO, ctypes?  Do they stay on the list or come off?

trotterdylan:
Off hand I'd say cPickle, cProfile and cStringIO should mirror their non-C counterparts. ctypes a bigger question, but ISTM we could implement it using cgo.

PR #221, opened 2017-01-27
Implement str mod width and fill

S-YOU:
- Supports `%<positive number>[rsfdxXo%]`
- 2 can be any number, huge numbers may panic or throw unexpected errors. (Python does not throw errors, but almost freeze as far as I tried. Or do we limit up to arbitrary number?)
- negative numbers are unsupported, will still throw errors same as existing one.
- decimal width on floating points is not supported, should throw same errors as before.
- moved zfill code outside as helper function, it is not only padding zero, so renamed to xfill.
- added python side tests and few go tests.


runtime/str.go
```diff
@@ -819,11 +819,17 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
 		if matches[7] != "%" && valueIndex >= len(values.elems) {
 			return nil, f.RaiseType(TypeErrorType, "not enough arguments for format string")
 		}
-		if matches[1] != "" {
-			return nil, f.RaiseType(NotImplementedErrorType, "conversion flags not yet supported")
-		}
-		if matches[2] != "" || matches[4] != "" {
-			return nil, f.RaiseType(NotImplementedErrorType, "field width not yet supported")
+		if (matches[1] == "" || matches[1] == "0") && matches[2] != "*" && (matches[7] == "d" || matches[7] == "x" || matches[7] == "X") {
```
trotterdylan: I think it's worth naming the various `matches[i]` that we use so that the code reads a little more clearly. E.g.

```
flags := matches[1]
fieldType := matches[7]
```

trotterdylan: You can invert this condition so that you don't need the empty if clause:

```
if (matches[1] != "" && matches[1] != "0") || matches[2] == "*" || (matches[7] != "d" && matches[7] != "x" && matches[7] != "X") {
```

runtime/str.go
```diff
@@ -871,6 +877,17 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
 					val = strings.ToUpper(val)
 				}
 			}
+			if matches[2] != "" {
+				width, raised := strconv.Atoi(matches[2])
```
S-YOU: I see, thats why it is called raised instead of err.

runtime/str.go
S-YOU: Updated, looks like my brain can't handle inverted expressions well :)

runtime/str.go
```diff
@@ -816,21 +816,26 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
...
 			return nil, f.RaiseType(TypeErrorType, "not enough arguments for format string")
 		}
-		if matches[1] != "" {
-			return nil, f.RaiseType(NotImplementedErrorType, "conversion flags not yet supported")
-		}
-		if matches[2] != "" || matches[4] != "" {
-			return nil, f.RaiseType(NotImplementedErrorType, "field width not yet supported")
+		if (flags != "" && flags != "0") || fieldWidth == "*" || (fieldType != "d" && fieldType != "x" && fieldType != "X") {
+			if flags != "" {
+				fmt.Println("matches1", flags, matches)
```
trotterdylan: Remove these debug statements

runtime/str.go
```diff
@@ -816,21 +816,26 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
...
+		flags, fieldWidth, fieldType := matches[1], matches[2], matches[7]
+		if fieldType != "%" && valueIndex >= len(values.elems) {
 			return nil, f.RaiseType(TypeErrorType, "not enough arguments for format string")
 		}
-		if matches[1] != "" {
-			return nil, f.RaiseType(NotImplementedErrorType, "conversion flags not yet supported")
-		}
-		if matches[2] != "" || matches[4] != "" {
-			return nil, f.RaiseType(NotImplementedErrorType, "field width not yet supported")
+		if (flags != "" && flags != "0") || fieldWidth == "*" || (fieldType != "d" && fieldType != "x" && fieldType != "X") {
```
trotterdylan: I would move the field width conversion out of the switch statement (since other types will use it eventually) and break up this large if condition a bit:

```
fieldWidth := -1
if matches[2] == "*" {
  return nil, f.RaiseType(...)
}
if matches[2] != "" {
  fieldWidth = .. Atoi conversion
}
```

Then in each field type block that doesn't support field width, raise when `fieldWidth != -1`. This will be more code but I think it makes it easier to reason about since each if statement has a very simple condition.

runtime/str.go
S-YOU: Oops, missed that. I have been a bit busy with other task in company. will remove.

S-YOU:
Reviews update, rebased and added supports for `r s f o %`

runtime/str.go
```diff
@@ -806,6 +806,8 @@ func strCompare(v, w *Object, ltResult, eqResult, gtResult *Int) *Object {
 }
 
 func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseException) {
+	var val string
+	var err error
```
trotterdylan: I would declare these closer to where they're used. Especially err, which only seems to be used in one block.

runtime/str.go
```diff
@@ -1093,3 +1112,21 @@ func toUpper(b byte) byte {
...
+	}
+	buf := bytes.Buffer{}
+	buf.Grow(width)
+	if l > 0 && fillchar == "0" && (s[0] == '-' || s[0] == '+') {
+		buf.WriteByte(s[0])
+		s = s[1:]
+		width--
+	}
+	// TODO: Support or throw fillchar len more than one.
+	buf.WriteString(strings.Repeat(fillchar, width-len(s)))
```
trotterdylan: Change `width-len(s)` to `width-l`

runtime/str.go
```diff
@@ -1093,3 +1112,21 @@ func toUpper(b byte) byte {
...
+		return s
+	}
+	buf := bytes.Buffer{}
+	buf.Grow(width)
+	if l > 0 && fillchar == "0" && (s[0] == '-' || s[0] == '+') {
+		buf.WriteByte(s[0])
+		s = s[1:]
+		width--
+	}
+	// TODO: Support or throw fillchar len more than one.
```
trotterdylan: I think it's worth mentioning in a comment that len(fillchar) == 1 is a precondition. E.g.:

```
// xfill returns s padded with fillchar so that its length is at least width. fillchar must be a single character.
```

Probably also worth mentioning the fillchar == "0" behavior.

runtime/str.go
```diff
@@ -1093,3 +1112,21 @@ func toUpper(b byte) byte {
 	}
 	return b
 }
+
+func xfill(s string, width int, fillchar string) string {
```
trotterdylan: I would call this strPad() or something a little more suggestive.

runtime/str.go
```diff
@@ -873,17 +894,28 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
...
+					fillchar = flags
+				}
+				val = xfill(val, fieldWidth, fillchar)
+			}
 			buf.WriteString(val)
 			valueIndex++
 		case "%":
-			buf.WriteString("%")
+			val = "%"
+			if fieldWidth > 0 {
```
trotterdylan: Is this the right behavior? I don't see this when I try it in an interpreter:

```
>>> '%8%'
'%8%'
```

runtime/str.go
S-YOU: len of s may change on above block.

runtime/str.go
S-YOU: Sure. to be more concise, I think it is a left pad. how about strLeftPad?

runtime/str.go
S-YOU: Make sense, thank you

runtime/str.go
S-YOU: there is no mod in your test, try with blank tuple `% tuple`

runtime/str.go
S-YOU: may I should update `l` on above block instead.

S-YOU:
Done review updates, and one comment above for `'%8%' % tuple()`

runtime/str.go
trotterdylan: Sounds good

runtime/str.go
trotterdylan: Heh that would explain the behavior I was seeing! Thanks for the correction.

S-YOU:
Changed strPad to strLeftPad.

runtime/str.go
S-YOU: Missed, your response, updated to strLeftPad now.

PR #222, opened 2017-01-27
Use stdlib unittest for md5 instead of custom one

S-YOU:
```
% cat third_party/stdlib/test/test_md5.py  | make run
test_basics (__main__.MD5_Test) ... ok
test_hexdigest (__main__.MD5_Test) ... ok
test_large_update (__main__.MD5_Test) ... ok

Ran 3 tests in 0.016706s

OK
```


Makefile
```diff
@@ -92,6 +92,7 @@ STDLIB_TESTS := \
   test/test_list \
   test/test_slice \
   test/test_string \
+	test/test_md5 \
```
trotterdylan: Use spaces here

S-YOU: ah, sorry

S-YOU: Done

PR #225, opened 2017-01-29
Implement list.remove

aisk:


runtime/list.go
```diff
@@ -170,6 +170,38 @@ func listCount(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	}
+	foundEqItem, raised := seqFindFirst(f, args[0], pred)
+	if raised != nil {
+		return nil, raised
+	}
+	if !foundEqItem {
+		return nil, f.RaiseType(ValueErrorType, "list.remove(x): x not in list")
+	}
+	l := toListUnsafe(args[0])
+	l.mutex.Lock()
```
trotterdylan: You need to lock the list during the seqFindFirst operation as well, otherwise the list could change between determining the item index and attempting to remove it.

aisk:
Done 😄

runtime/list.go
```diff
@@ -170,6 +170,38 @@ func listCount(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+			return false, raised
+		}
+		ret, raised := IsTrue(f, eq)
+		if raised != nil {
+			return false, raised
+		}
+		return ret, nil
+	}
+	l := toListUnsafe(args[0])
+	l.mutex.RLock()
```
trotterdylan: An RLock is insufficient since you're actually going to modify the slice. This should be Lock()

runtime/list.go
```diff
@@ -170,6 +170,38 @@ func listCount(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	defer l.mutex.RUnlock()
```
trotterdylan: Avoid defer except in cases where performance doesn't matter. There is a huge tax to using defer. It may help to structure your code like this to avoid doing multiple unlocks:

```
foundEqItem, raised := seqFindFirst(f, args[0], pred)
if raised == nil {
  if foundEqItem {
    raised = f.RaiseType(ValueErrorType, "list.remove(x): x not in list")
  } else {
    l.elems = append(l.elems[:i], l.elems[i+1:]...)
  }
}
l.mutex.Unlock()
if raised != nil {
  return nil, raised
}
return None, nil
```

runtime/list_test.go
```diff
@@ -92,6 +92,29 @@ func TestListCount(t *testing.T) {
...
+		}
+		if _, raised := rem.Call(f, args, nil); raised != nil {
+			return nil, raised
+		}
+		return args[0], nil
+	}).ToObject()
+	cases := []invokeTestCase{
+		{args: wrapArgs(newTestList(1, 2, 3), 2), want: newTestList(1, 3).ToObject()},
+		{args: wrapArgs(NewList()), wantExc: mustCreateException(TypeErrorType, "'remove' of 'list' requires 2 arguments")},
+		{args: wrapArgs(NewList(), 1), wantExc: mustCreateException(ValueErrorType, "list.remove(x): x not in list")},
```
trotterdylan: Could you also add a test case where the sought element is in the list more than once?

cclauss: Deferred performance doubled in V1.8 with more speedups expected in v1.9  https://github.com/golang/go/issues/14939#issuecomment-252935757

trotterdylan: That's true but we also target 1.6 and 1.7 :)

Additionally, in my experience, even a 4x improvement would not be sufficient to justify its use in low level code. The underlying issue is that it does an allocation in most cases. Much of Grumpy's design and many optimizations revolve around avoiding allocations because they're so expensive.

runtime/list.go
aisk: Looks like we got a deadlock while using Lock()

trotterdylan: Ah, that's because seqFindFirst uses an iterator which does an RLock internally (see `listIteratorNext`). It's probably simpler anyway to do a Lock and then iterate over the elems slice directly.

aisk: Got

aisk:
Updated.

runtime/list.go
```diff
@@ -170,6 +170,40 @@ func listCount(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+		return nil, raised
+	}
+	l := toListUnsafe(args[0])
+	l.mutex.Lock()
+	value := args[1]
+	found := false
+	var raised *BaseException
+	for i, elem := range l.elems {
+		var eq *Object
+		eq, raised = Eq(f, elem, value)
```
trotterdylan: Minor style suggestion: put these assignments directly in the if statement:

```
if eq, raised = Eq(f, elem, value); raised != nil {
```

Ditto for the next if statement.

runtime/list.go
```diff
@@ -170,6 +170,40 @@ func listCount(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+		if found {
+			l.elems = append(l.elems[:i], l.elems[i+1:]...)
+			break
+		}
+	}
+	l.mutex.Unlock()
+	if raised != nil {
+		return nil, raised
+	}
+	if found {
```
trotterdylan: I would do the `if !found {` condition here so that the default return at the bottom of the function is "success".

runtime/list.go
aisk: I mixed up Rlock with Recursive lock, looks like it's much handy if go had this stuff.

runtime/list.go
aisk: Updated.

runtime/list.go
trotterdylan: Recursive mutex is a [contentious topic](https://groups.google.com/forum/#!topic/golang-nuts/vN5ncBdtkcA) in the Go community. It was intentionally left out because they can be error prone.

As it happens, it's not possible to implement some Python functionality without recursive locks, so I've had to [implement](https://github.com/google/grumpy/blob/master/runtime/threading.go#L49) a cheesy recursive lock for Grumpy. This is likely also how we'll implement the Python threading.RLock class.

PR #227, opened 2017-01-31
Implement __hex__ and __oct__

choleraehyq:
* Implement default `__hex__` and `__oct__` for `int` and `long`
* Refactor builtin `hex` and `oct`.

Signed-off-by: choleraehyq <choleraehyq@gmail.com>

### DIFF
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -253,6 +253,29 @@ func Hash(f *Frame, o *Object) (*Int, *BaseException) {
 	return toIntUnsafe(h), nil
 }
 
+// Hex returns the result of o.__hex__ if defined.
+func Hex(f *Frame, o *Object) (*Object, *BaseException) {
+	hex := o.typ.slots.Hex
+	if hex == nil {
+		if method, raised := o.typ.mroLookup(f, NewStr("__hex__")); raised != nil {
+			return nil, raised
+		} else if method != nil {
+			return method.Call(f, nil, nil)
+		} else {
+			_, raised := hexNotImplemented(f, o)
+			return nil, raised
+		}
+	}
+	h, raised := hex.Fn(f, o)
+	if raised != nil {
+		return nil, raised
+	}
+	if !h.isInstance(StrType) {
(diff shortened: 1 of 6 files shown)


runtime/core.go
```diff
@@ -253,6 +253,29 @@ func Hash(f *Frame, o *Object) (*Int, *BaseException) {
 	return toIntUnsafe(h), nil
 }
 
+// Hex returns the result of o.__hex__ if defined.
+func Hex(f *Frame, o *Object) (*Object, *BaseException) {
+	hex := o.typ.slots.Hex
+	if hex == nil {
+		if method, raised := o.typ.mroLookup(f, NewStr("__hex__")); raised != nil {
+			return nil, raised
+		} else if method != nil {
```
trotterdylan: Remove redundant `else`. Prefer this form:

```
if ... {
  return ...
}
if ... {
  return ...
}
```

runtime/core.go
```diff
@@ -253,6 +253,29 @@ func Hash(f *Frame, o *Object) (*Int, *BaseException) {
 	return toIntUnsafe(h), nil
 }
 
+// Hex returns the result of o.__hex__ if defined.
+func Hex(f *Frame, o *Object) (*Object, *BaseException) {
+	hex := o.typ.slots.Hex
+	if hex == nil {
+		if method, raised := o.typ.mroLookup(f, NewStr("__hex__")); raised != nil {
```
trotterdylan: No need to lookup `__hex__`. If that method exists, an automatically generated Hex slot will exist and forward to `__hex__`.

runtime/core.go
```diff
@@ -1181,6 +1227,14 @@ func hashNotImplemented(f *Frame, o *Object) (*Object, *BaseException) {
 	return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("unhashable type: '%s'", o.typ.Name()))
 }
 
+func hexNotImplemented(f *Frame, o *Object) (*Object, *BaseException) {
```
trotterdylan: Prefer to do this raise inline rather than as a helper. Use helpers only when there is a significant amount of code being factored out or there is a lot of duplication.

choleraehyq:
@trotterdylan Done!

PR #229, opened 2017-02-01
The implementation of str.swapcase() operation.

corona10:
Hi, this is my first PR for your project.
and please give me feedbacks if possible.
And is there any way to run a specific unit test after I modify a specific unit test?

PTAL.

### DIFF
--- a/runtime/str.go
+++ b/runtime/str.go
@@ -748,6 +748,29 @@ func strStr(_ *Frame, o *Object) (*Object, *BaseException) {
 	return NewStr(toStrUnsafe(o).Value()).ToObject(), nil
 }
 
+func strSwapCase(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	expectedTypes := []*Type{StrType}
+	if raised := checkMethodArgs(f, "swapcase", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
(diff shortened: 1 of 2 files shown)


corona10:
For reviewers, I found that an 1 test was written in mistake.
So I fixed it and run on my Travis CI and all tests were passed.

runtime/str.go
```diff
@@ -748,6 +748,29 @@ func strStr(_ *Frame, o *Object) (*Object, *BaseException) {
 	return NewStr(toStrUnsafe(o).Value()).ToObject(), nil
 }
 
+func strSwapCase(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	expectedTypes := []*Type{StrType}
+	if raised := checkMethodArgs(f, "swapcase", args, expectedTypes...); raised != nil {
```
trotterdylan: You can pass Strtype directly as an argument rather than expanding the expectedTypes slice this way:

```
if raised := checkMethodArgs(f, "swapcase", args, StrType); raised != nil {
```

corona10: @trotterdylan 
Thank you for reviewing my codes.
I changed it! PTAL.

corona10:
@trotterdylan
Thank you for reviewing my codes.
I changed it! PTAL.

PR #230, opened 2017-02-02
Implement Ellipsis type 

corona10:
I found that an ellipsis type is not yet implemented.
So I implement and add unit test code for it.
Please take a look and give me feedbacks.


corona10:
@trotterdylan It's just like numeric extensions for python. It can be an element of python's list, tuple, slice. As you mentioned it looks like to used in the numpy and the scipy.

PR #231, opened 2017-02-02
Implement list.index

aisk:
~~Sorry, forgot the lock, plz don't review now.~~
Ready for review.

### DIFF
--- a/runtime/list.go
+++ b/runtime/list.go
@@ -174,30 +174,20 @@ func listRemove(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	if raised := checkMethodArgs(f, "remove", args, ListType, ObjectType); raised != nil {
 		return nil, raised
 	}
+	value := args[1]
 	l := toListUnsafe(args[0])
 	l.mutex.Lock()
-	value := args[1]
-	found := false
-	var raised *BaseException
-	for i, elem := range l.elems {
-		var eq *Object
-		if eq, raised = Eq(f, elem, value); raised != nil {
-			break
-		}
-		if found, raised = IsTrue(f, eq); raised != nil {
-			break
-		}
-		if found {
-			l.elems = append(l.elems[:i], l.elems[i+1:]...)
-			break
+	index, raised := seqFindElem(f, l.elems, value)
+	if raised == nil {
+		if index != -1 {
+			l.elems = append(l.elems[:index], l.elems[index+1:]...)
+		} else {
+			raised = f.RaiseType(ValueErrorType, "list.remove(x): x not in list")
 		}
 	}
 	l.mutex.Unlock()
 	if raised != nil {
 		return nil, raised
-	}
-	if !found {
-		return nil, f.RaiseType(ValueErrorType, "list.remove(x): x not in list")
 	}
 	return None, nil
 }
@@ -338,6 +328,50 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return listCompare(f, toListUnsafe(v), w, NE)
 }
 
+func listIndex(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	expectedTypes := []*Type{ListType, ObjectType, ObjectType, ObjectType}
+	argc := len(args)
+	var raised *BaseException
+	if argc == 2 || argc == 3 {
+		expectedTypes = expectedTypes[:argc]
+	}
+	if raised = checkMethodArgs(f, "index", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	l := toListUnsafe(args[0])
+	l.mutex.RLock()
+	numElems := len(l.elems)
+	start, stop := 0, numElems
+	if argc > 2 {
+		start, raised = IndexInt(f, args[2])
+		if raised != nil {
+			l.mutex.RUnlock()
+			return nil, raised
+		}
+	}
+	if argc > 3 {
+		stop, raised = IndexInt(f, args[3])
+		if raised != nil {
+			l.mutex.RUnlock()
+			return nil, raised
+		}
+	}
+	start, stop = adjustIndex(start, stop, numElems)
+	value := args[1]
+	index := -1
+	if start < numElems && start < stop {
+		index, raised = seqFindElem(f, l.elems[start:stop], value)
+	}
+	l.mutex.RUnlock()
+	if raised != nil {
+		return nil, raised
+	}
+	if index == -1 {
+		return nil, f.RaiseType(ValueErrorType, fmt.Sprintf("%v is not in list", value))
+	}
+	return NewInt(index + start).ToObject(), nil
+}
+
 func listPop(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	argc := len(args)
 	expectedTypes := []*Type{ListType, ObjectType}
@@ -432,6 +466,7 @@ func initListType(dict map[string]*Object) {
 	dict["append"] = newBuiltinFunction("append", listAppend).ToObject()
 	dict["count"] = newBuiltinFunction("count", listCount).ToObject()
 	dict["extend"] = newBuiltinFunction("extend", listExtend).ToObject()
+	dict["index"] = newBuiltinFunction("index", listIndex).ToObject()
 	dict["insert"] = newBuiltinFunction("insert", listInsert).ToObject()
 	dict["pop"] = newBuiltinFunction("pop", listPop).ToObject()
 	dict["remove"] = newBuiltinFunction("remove", listRemove).ToObject()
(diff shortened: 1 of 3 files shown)


runtime/list.go
```diff
@@ -306,6 +306,70 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return listCompare(f, toListUnsafe(v), w, NE)
 }
 
+func listIndex(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	expectedTypes := []*Type{ListType, ObjectType, IntType, IntType}
```
trotterdylan: I think that for the last two parameters, index() accepts any object that implements `__index__`. See for example [strIndex()](https://github.com/google/grumpy/blob/master/runtime/str.go#L238) which calls IndexInt() to pull out an integer index from an arbitrary Object.

runtime/list.go
```diff
@@ -306,6 +306,70 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	}
+	if raised := checkMethodArgs(f, "index", args, expectedTypes...); raised != nil {
+		l.mutex.RUnlock()
+		return nil, raised
+	}
+	value := args[1]
+	index := 0
+	found := false
+	var raised *BaseException
+	for i := start; i < stop; i++ {
```
trotterdylan: This seems very similar to the implementation of remove() so maybe it could be factored out into a helper function in seq.go that takes a slice and returns the index.

aisk:
WIP.

aisk:
Updated, but don't refactor the common codes in list.index / list.remove.

I modified the seqFindFisrt, and let it returns the first position where it found the element, but got a deadlock, like which we found in other PR.

Should we add a new function, and don't convert it to a seq, or just keep them just like now?

trotterdylan:
Right, I was thinking that you'd need a new function in seq.go, something like:

```
// seqFindElem returns the index of the element in elems that is equal to o, or -1 if no element is equal.
seqFindElem(f *Frame, elems []*Object, o *Object) (int, *BaseException)
```

aisk:
Updated.

runtime/seq.go
```diff
@@ -156,6 +156,7 @@ func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) (bool, *BaseExc
 	if raised != nil {
 		return false, raised
 	}
+	index := 0
```
trotterdylan: I think this is leftover from some obsolete changes?

runtime/list.go
```diff
@@ -338,6 +338,72 @@ func listNE(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	l := toListUnsafe(args[0])
+	var start, stop int
+	l.mutex.RLock()
+	switch argc {
+	case 2:
+		start = 0
+		stop = len(l.elems)
+	case 3:
+		numElems := len(l.elems)
+		start, raised = IndexInt(f, args[2])
```
trotterdylan: I think you can eliminate a bit of code duplication if you rearrange this as repeated if statements rather than a switch:

```
start, stop := 0, len(l.elems)
if argc >= 2 {
  start, raised = IndexInt(f, args[2])
  ...
}
if argc >= 3 {
  stop, raised = IndexInt(f, args[3])
  ...
}
```

aisk:
Updated.

runtime/seq.go
```diff
@@ -173,6 +173,24 @@ func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) (bool, *BaseExc
 	return false, nil
 }
 
+func seqFindElem(f *Frame, elems []*Object, o *Object, start, stop int) (int, bool, *BaseException) {
```
trotterdylan: Rather than passing start and stop, I would have the caller slice the elems: `seqFindElems(f, elems[start:stop], o)`

trotterdylan: Have the return value be `(int, *BaseException)` and have the first return value be -1 when not found.

runtime/list_test.go
```diff
@@ -91,6 +91,35 @@ func TestListCount(t *testing.T) {
...
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 0, 4), want: NewInt(3).ToObject()},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 0, 3), wantExc: mustCreateException(ValueErrorType, "3 is not in list")},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, -2), want: NewInt(3).ToObject()},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, -1), wantExc: mustCreateException(ValueErrorType, "3 is not in list")},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 0, -1), want: NewInt(3).ToObject()},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 0, -2), wantExc: mustCreateException(ValueErrorType, "3 is not in list")},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 0, 999), want: NewInt(3).ToObject()},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), "foo", 0, 999), wantExc: mustCreateException(ValueErrorType, "'foo' is not in list")},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 999), wantExc: mustCreateException(ValueErrorType, "3 is not in list")},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 5, 0), wantExc: mustCreateException(ValueErrorType, "3 is not in list")},
```
trotterdylan: Can you add some test cases with repeated values and also some values that aren't integers?

aisk:
Ready for review again.

runtime/list_test.go
trotterdylan: Ping, can you address this request?

runtime/seq.go
```diff
@@ -173,6 +173,25 @@ func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) (bool, *BaseExc
 	return false, nil
 }
 
+func seqFindElem(f *Frame, elems []*Object, o *Object) (int, *BaseException) {
+	// for i := start; i < stop; i++ {
```
trotterdylan: remove this commented out code

runtime/seq.go
```diff
@@ -173,6 +173,25 @@ func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) (bool, *BaseExc
...
+	for i, elem := range elems {
+		var raised *BaseException
+		var eq *Object
+		var found bool
+		if eq, raised = Eq(f, elem, o); raised != nil {
```
trotterdylan: Slightly less verbose:

```
eq, raised := Eq(f, elem, o)
if raised != nil {
  return -1, raised
}
found, raised := IsTrue(f, eq)
if raised != nil {
  return -1, raised
}
```

runtime/list_test.go
```diff
@@ -91,6 +91,36 @@ func TestListCount(t *testing.T) {
...
+		"__index__": newBuiltinFunction("__index__", func(f *Frame, _ Args, _ KWArgs) (*Object, *BaseException) {
+			return NewInt(0).ToObject(), nil
+		}).ToObject(),
+	}))
+	cases := []invokeTestCase{
+		// {args: wrapArgs(newTestList(), 1, "foo"), wantExc: mustCreateException(TypeErrorType, "slice indices must be integers or None or have an __index__ method")},
+		{args: wrapArgs(newTestList(10, 20, 30), 20), want: NewInt(1).ToObject()},
+		{args: wrapArgs(newTestList(10, 20, 30), 20, newObject(intIndexType)), want: NewInt(1).ToObject()},
+		{args: wrapArgs(newTestList(0, 1, 2, 3, 4), 3, 3), want: NewInt(3).ToObject()},
+		{args: wrapArgs(newTestList(0, 2.0, 2, 3, 4, 2, 1, "foo"), 3, 3), want: NewInt(3).ToObject()},
```
aisk: @trotterdylan Hi, the repeated values test case is here.

aisk:
Updated.

runtime/seq.go
```diff
@@ -173,6 +173,26 @@ func seqFindFirst(f *Frame, iterable *Object, pred func(*Object) (bool, *BaseExc
 	return false, nil
 }
 
+func seqFindElem(f *Frame, elems []*Object, o *Object) (int, *BaseException) {
+	for i, elem := range elems {
+		var raised *BaseException
+		var eq *Object
+		var found bool
+		eq, raised = Eq(f, elem, o)
```
trotterdylan: You can remove the eq and raised declarations above and then assign them this way:

```
eq, raised := Eq(f, elem, o)
```

Ditto for found, below.

PR #233, opened 2017-02-03
Implementation of raw_input()

corona10:
I implemented a simplified version of raw_input().

Please review on focusing these.
First, It might be not 100% compatible with CPython and pypy's raw_input(). 
and also need more implementation to handle error cases. 

Second, I need some guide for writing unit test codes to check it is well working for raw_input().
I don't know how to test user's input from keyboard.

Third, prompt printing is slower when running with `go run *.go` so if you have solutions then please give me a feedback.

Related PR= #247


runtime/builtin_types.go
```diff
@@ -559,6 +560,42 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
 }
 
+func builtinRawInput(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	if len(args) > 1 {
+		msg := fmt.Sprintf("[raw_]input expcted at most 1 arguments, got %d", len(args))
+		return nil, f.RaiseType(TypeErrorType, msg)
+	}
+
+	fin := os.Stdin
+	fout := os.Stdout
```
trotterdylan: Rather than writing to Stdout directly, you could use the Print() function: `Print(f, args, false)`

runtime/builtin_types.go
```diff
@@ -559,6 +560,42 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
+		msg := fmt.Sprintf("[raw_]input: lost sys.stdout")
+		return nil, f.RaiseType(RuntimeErrorType, msg)
+	}
+
+	in := bufio.NewReader(fin)
+	if len(args) == 1 {
+		prompt, _ := ToStr(f, args[0])
+		fmt.Fprint(fout, prompt.Value())
+	}
+	s, _, _ := in.ReadLine()
```
trotterdylan: ReadLine() isn't ideal here because it can return with less than a full line. I think the right thing to do is probably use the grumpy.File object representing sys.stdin and call readLine() on it. But there's no clear way to get the sys module from here.

One possible solution that would be to create Stdin, Stdout and Stderr variables in the grumpy package that wrap os.Stdin/out/err with a grumpy.File object, and then have the sys module expose those instead of creating new files. The grumpy.Print() function could then use the new Stdout object. This isn't perfect but I think it is an improvement over what we have today.

runtime/builtin_types.go
```diff
@@ -559,6 +560,42 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
+
+	// Todo: Should implement more cases of error.
+	pySsizeTmax := ((^uint(0)) - 1) >> 1
```
trotterdylan: I think that if len(s) were ever > than this value that Go would panic so don't bother checking this.

runtime/builtin_types.go
```diff
@@ -559,6 +560,42 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
+	size := len(s)
+	if uint(size) > pySsizeTmax {
+		msg := fmt.Sprintf("[raw_]input: input too long")
+		return nil, f.RaiseType(OverflowErrorType, msg)
+	}
+
+	return NewStr(string(s[0:size])).ToObject(), nil
```
trotterdylan: How is this different than `NewStr(s)`?

corona10:
@trotterdylan Thank you for great and awesome review. I will try to fix it.

corona10:
@trotterdylan Updated! I fixed codes as you commented. Actually, about using grumpy.File object is not 100% same as you intended. I found more simpler way.
And I add unit test code, but it should be improvement later time. I wish other guys could improve this unit test to check user's input and function's output. Unfortunately in now time, this is out of my ability because I learned go lang just few days ago.

runtime/builtin_types.go
```diff
@@ -559,6 +560,33 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
+	if stdout == nil {
+		msg := fmt.Sprintf("[raw_]input: lost sys.stdout")
+		return nil, f.RaiseType(RuntimeErrorType, msg)
+	}
+
+	if len(args) == 1 {
+		pyPrint(f, args, "", "", stdout)
+	}
+
+	line, _ := stdin.readLine(MaxInt)
```
trotterdylan: I think I was wrong about implementing it this way :(

The trouble is that the C implementation uses fgets on the underlying C file pointer. fgets reads until it reaches `\n` but readLine will stop when it reaches `\r` and will convert that to a `\n`.

Probably the simplest thing to do here is to call `stdin.reader.ReadString("\n")` instead. This would mean you'd have to handle the EOF error case, however.

trotterdylan: You must handle the error case. It's pretty simple to do:

```
if err != nil {
  return nil, f.RaiseType(IOErrorType, err.Error())
}
```

runtime/builtin_types.go
```diff
@@ -559,6 +560,33 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
 }
 
+func builtinRawInput(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	if len(args) > 1 {
+		msg := fmt.Sprintf("[raw_]input expcted at most 1 arguments, got %d", len(args))
+		return nil, f.RaiseType(TypeErrorType, msg)
+	}
+
+	stdout := os.Stdout
+	stdin := NewFileFromFD(os.Stdin.Fd())
```
trotterdylan: The trouble with creating a new File wrapping os.Stdin is that it will have a different buffered reader than the one held by sys.stdin. This is especially problematic because this File object is cleaned up after builtinRawInput() returns, meaning that whatever buffered data it held just disappears.

This goes back to my suggestion of having grumpy.Stdin/out/err variables that get used by this function and having sys.stdin/out/err be aliases for those variables:

```
# sys.py
from __go__.grumpy import Stdin as stdin, Stdout as stdout, Stderr as stderr
```

This way any buffering done for one call will be used by subsequent calls.

runtime/builtin_types_test.go
```diff
@@ -422,3 +422,24 @@ func TestBuiltinSetAttr(t *testing.T) {
 		}
 	}
 }
+
+// TODO: Improvment unit tests codes into catch user's input and output.
```
trotterdylan: I'd want to see some additional tests that exercise the whole builtinRawInput() function before I approve this PR. It's important that new code is fully tested because the probability that somebody retroactively adds tests is low.

corona10:
@trotterdylan A lot of stuff!! I will do it as your comment. Thanks

corona10:
@trotterdylan Updated!! Please take a look.

runtime/builtin_types.go
```diff
@@ -559,6 +560,35 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
 	return ListType.Call(f, []*Object{r}, nil)
 }
 
+func builtinRawInput(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	if len(args) > 1 {
+		msg := fmt.Sprintf("[raw_]input expcted at most 1 arguments, got %d", len(args))
+		return nil, f.RaiseType(TypeErrorType, msg)
+	}
+
+	osStdout := os.Stdout
```
trotterdylan: Use the new grumpy.Stdout variable for output. This will require modifying pyPrint(), Print(), etc. to take a *grumpy.File instead of *os.File. Might be better to add Stdin/out/err and change the print behavior in a separate PR and once that's merged, continue with this one.

runtime/builtin_types_test.go
```diff
@@ -422,3 +423,63 @@ func TestBuiltinSetAttr(t *testing.T) {
 		}
 	}
 }
+
+func TestRawInput(t *testing.T) {
+	fun := newBuiltinFunction("TestRawInput", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		// Create a fake Stdin for input test.
+		stdinFile, _ := ioutil.TempFile(os.TempDir(), "prefix")
+		stdinFile.Write([]byte("hello grumpy\n"))
+		stdinFile.Close()
```
trotterdylan: I don't think you need to reopen stdinFile. You should be able to read what you've written. You may need to Sync() (and possibly Seek() to the beginning, but I don't think so).

runtime/builtin_types_test.go
```diff
@@ -422,3 +423,63 @@ func TestBuiltinSetAttr(t *testing.T) {
 		}
 	}
 }
+
+func TestRawInput(t *testing.T) {
+	fun := newBuiltinFunction("TestRawInput", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		// Create a fake Stdin for input test.
+		stdinFile, _ := ioutil.TempFile(os.TempDir(), "prefix")
```
trotterdylan: You need to check errors for these io operations.

trotterdylan: Not sure if it'd be easier, but you could also use a Pipe() instead of a temp file for stdin, similar to what you've done for stdout. Something like this:

```
stdinFile, w := os.Pipe()
go func() {
  w.Write([]byte("hello grumpy\n"))
  w.Close()
}()
```

You need to do the write asynchronously this way in case the Write() call blocks waiting for a Read() on the other end of the pipe.

runtime/builtin_types_test.go
```diff
@@ -422,3 +423,63 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+			return nil, f.RaiseType(RuntimeErrorType, fmt.Sprintf("failed to open pipe: %v", err))
+		}
+
+		oldStdout := os.Stdout
+		os.Stdout = w
+		defer func() {
+			os.Stdout = oldStdout
+			Stdin = oldStdin
+			stdinFile.Close()
+			os.Remove(stdinFile.Name())
```
trotterdylan: Move `defer os.Remove(stdinFile.Name())` up to where you create the file so that an early return won't leave the tempfile lying around.

runtime/builtin_types_test.go
```diff
@@ -422,3 +423,63 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+
+		if _, err := io.Copy(&output, r); err != nil {
+			return nil, f.RaiseType(RuntimeErrorType, fmt.Sprintf("failed to open pipe: %v", err))
+		}
+		<-done
+
+		if raised != nil {
+			return nil, raised
+		}
+		result := newTestTuple(input, output.String()).ToObject()
```
trotterdylan: `return newTestTuple(input, output.String()).ToObject(), nil`

runtime/builtin_types_test.go
corona10: Cool~! I will try it!!

corona10:
@trotterdylan Updated. PTAL

runtime/builtin_types.go
```diff
@@ -559,6 +560,34 @@ func builtinRange(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException)
...
+		return nil, f.RaiseType(RuntimeErrorType, msg)
+	}
+
+	if Stdout == nil {
+		msg := fmt.Sprintf("[raw_]input: lost sys.stdout")
+		return nil, f.RaiseType(RuntimeErrorType, msg)
+	}
+
+	if len(args) == 1 {
+		pyPrint(f, args, "", "", Stdout)
```
trotterdylan: Check the resulting exception

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,72 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+func TestRawInput(t *testing.T) {
+	fun := newBuiltinFunction("TestRawInput", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		// Create a fake Stdin for input test.
+		stdinFile, w, err := os.Pipe()
+		if err != nil {
+			return nil, f.RaiseType(RuntimeErrorType, fmt.Sprintf("failed to open pipe: %v", err))
+		}
+
+		go func() {
+			defer w.Close()
```
trotterdylan: Don't bother defering, just Close after the Write

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,72 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+			w.Write([]byte("hello grumpy\n"))
+			stdinFile.Sync()
```
trotterdylan: It's w that needs to Sync, but closing it will do the same thing.

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,72 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+
+		oldStdin := Stdin
+		Stdin = NewFileFromFD(stdinFile.Fd())
+		defer func() {
+			Stdin = oldStdin
+			stdinFile.Close()
+		}()
+
+		// Create a fake Stdout for output test.
+		r, stdoutFile, err := os.Pipe()
```
trotterdylan: Any way we can use captureStdout for this part?

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,72 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+		defer func() {
+			Stdout = oldStdout
+		}()
+		done := make(chan struct{})
+		var input *Object
+		var output bytes.Buffer
+		var raised *BaseException
+
+		go func() {
+			defer close(done)
```
trotterdylan: No need to defer either of these, just call them normally below

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,72 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+		}
+
+		oldStdout := Stdout
+		Stdout = NewFileFromFD(stdoutFile.Fd())
+		defer func() {
+			Stdout = oldStdout
+		}()
+		done := make(chan struct{})
+		var input *Object
+		var output bytes.Buffer
```
trotterdylan: Declare this where it's used down below when passed to Copy

runtime/builtin_types_test.go
corona10: @trotterdylan I want to do by the same way you mentioned. But until we refactoring `captureStdout`, I think not possible right now.

corona10: If we could change this [part](https://github.com/google/grumpy/blob/master/runtime/builtin_types_test.go#L363) for general functions. It might be possible.

corona10:
@trotterdylan Updated. 

p.s Can I get good materials to know when to use `defer`? I'm just using it as `finally` statement.

runtime/builtin_types_test.go
trotterdylan: What specifically needs to be generalized? Any inputs or outputs can made part of the closure for the passed function and read or written inside the function call.

trotterdylan:
@corona10 re: defer: It can be used like finally. But, like finally, it's only really useful if there are multiple possible paths for exiting the function or an exception (panic in Go) is raised. In this case, a panic is a fatal error so it doesn't really matter if we close things. The only time it's really important to run some code in the case of panic is if the panic may be handled somewhere (for the most part in the Grumpy code, we assume panic is fatal) or if you need to release some resource even in the event of a panic. Sometimes mutex falls into this category because it can otherwise cause a deadlock if not released. It's also good to guarantee that temp files are cleaned up. But otherwise, the possibility of panic is not usually a good reason to defer.

runtime/builtin_types_test.go
corona10: @trotterdylan In this case, `builtinRawInput()` will return `*Object `and `*BaseException`, but `fn` in `captureStdout` want to return only `*BaseException`. It will works for any function I will pass?

trotterdylan: I'm probably oversimplifying, but would something like this work?

```
var input *Object
raised := captureStdout(func() *BaseException {
  input = builtinRawInput(...)
})
if raised != nil {
  ...
}
// validate input
```

corona10: @trotterdylan Awesome! You are right!! Thank you for guiding me.  👍

corona10:
@trotterdylan Thank you for review and great guide! I updated it and please take a look. Thanks to you, I learned about how Go works.

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,51 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+	fun := newBuiltinFunction("TestRawInput", func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		// Create a fake Stdin for input test.
+		stdinFile, w, err := os.Pipe()
+		if err != nil {
+			return nil, f.RaiseType(RuntimeErrorType, fmt.Sprintf("failed to open pipe: %v", err))
+		}
+
+		go func() {
+			w.Write([]byte("hello grumpy\n"))
+			w.Sync()
```
trotterdylan: I think you want to close w here since that's the end of stdin.

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,51 @@ func TestBuiltinSetAttr(t *testing.T) {
...
+
+		if raised != nil {
+			return nil, raised
+		}
+
+		return newTestTuple(input, output).ToObject(), nil
+	}).ToObject()
+	cases := []invokeTestCase{
+		{args: wrapArgs(), want: newTestTuple("hello grumpy", "").ToObject()},
+		{args: wrapArgs("abc"), want: newTestTuple("hello grumpy", "abc").ToObject()},
```
trotterdylan: I think there's a few more things that need testing:

- Multiline prompt parameter: make sure all string lines are printed
- Multiple lines on stdin: only the first line should be returned by the raw_input call
- An empty stdin should raise EOFError:

```
$ python -c 'raw_input()' < /dev/null
  File "<string>", line 1, in <module>
EOFError: EOF when reading a line
```

The last two test cases can be supported by having `fun` take a parameter that is the string to put on stdin.

corona10:
@trotterdylan Updated!! PTAL

runtime/builtin_types_test.go
```diff
@@ -425,6 +425,69 @@ func TestBuiltinSetAttr(t *testing.T) {
 	}
 }
 
+func TestRawInput(t *testing.T) {
+	inputTest := func(s string, cases []invokeTestCase) {
```
trotterdylan: This organization seems like a reasonable way to support an additional parameter, but it's a little unorthodox. The way I've usually done it is something like this:

```
fun := wrapFuncForTest(func(f *Frame, s string, args ...*Object) (*Object, *BaseException) {
  ...
  go func() {
    w.Write([]byte(s))
    w.Close()
  }()
  ...
    in, raised := builtinRawInput(f, args, nil)
  ...
})
...
    {args: wrapArgs("hello\n", "show"), want: newTestTuple("hello", "show").ToObject()},
```

corona10:
@trotterdylan Thank you for reviewing long days :-)

pombredanne:
@corona10 very nice :)

PR #234, opened 2017-02-04
Adding bisect module from stdlib

GarryBullock:
Commented out the section that tries to import the c version.


trotterdylan:
Thanks for adding the tests. Merging.

PR #237, opened 2017-02-05
Implement dict.pop

aisk:
Just added this method by simply call `dict.GetItem` and `dict.DelItem`. But looks like it will lookup table twice.

I don't know this is acceptable, or should i change it to operate the dict structure directly?

### DIFF
--- a/runtime/dict.go
+++ b/runtime/dict.go
@@ -326,25 +326,11 @@ func (d *Dict) incVersion() {
 // DelItem removes the entry associated with key from d. It returns true if an
 // item was removed, or false if it did not exist in d.
 func (d *Dict) DelItem(f *Frame, key *Object) (bool, *BaseException) {
-	hash, raised := Hash(f, key)
+	originValue, raised := d.putItem(f, key, nil)
 	if raised != nil {
 		return false, raised
 	}
-	deleted := false
-	d.mutex.Lock(f)
-	v := d.version
-	if index, entry, raised := d.table.lookupEntry(f, hash.Value(), key); raised == nil {
-		if v != d.version {
-			raised = f.RaiseType(RuntimeErrorType, "dictionary changed during write")
-		} else if entry != nil && entry != deletedEntry {
-			d.table.storeEntry(index, deletedEntry)
-			d.table.incUsed(-1)
-			d.incVersion()
-			deleted = true
-		}
-	}
-	d.mutex.Unlock(f)
-	return deleted, raised
+	return originValue != nil, nil
 }
 
 // DelItemString removes the entry associated with key from d. It returns true
@@ -376,6 +362,12 @@ func (d *Dict) GetItemString(f *Frame, key string) (*Object, *BaseException) {
 	return d.GetItem(f, NewStr(key).ToObject())
 }
 
+// Pop looks up key in d, returning and removing the associalted value if exist,
+// or nil if key is not present in d.
+func (d *Dict) Pop(f *Frame, key *Object) (*Object, *BaseException) {
+	return d.putItem(f, key, nil)
+}
+
 // Keys returns a list containing all the keys in d.
 func (d *Dict) Keys(f *Frame) *List {
 	d.mutex.Lock(f)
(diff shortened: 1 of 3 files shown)


runtime/dict.go
```diff
@@ -624,6 +624,34 @@ func dictKeys(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	item, raised := d.GetItem(f, key)
+	if raised == nil {
+		if item == nil {
+			if argc > 2 {
+				item = args[2]
+			} else {
+				raised = raiseKeyError(f, key)
+			}
+		} else {
+			_, raised = d.DelItem(f, key)
```
trotterdylan: When DelItem() raises, both item and raised will be non-nil when returned below. Some callers may assume that item != nil implies raised == nil.

runtime/dict.go
```diff
@@ -624,6 +624,34 @@ func dictKeys(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	if argc == 2 {
+		expectedTypes = expectedTypes[:2]
+	}
+	if raised := checkMethodArgs(f, "pop", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	key := args[1]
+	d := toDictUnsafe(args[0])
+	d.mutex.Lock(f)
+	item, raised := d.GetItem(f, key)
```
trotterdylan: As you mention, calling GetItem then DelItem requires multiple lookups which includes hashing, equality check, etc. It'd be better to call lookupEntry() directly so that you can then remove it without having to look it up again.

One possible approach would be to modify putItem() so that it works when passed value == nil. Then, rather than returning a bool, have it return the original entry that was present (or nil if entry == deletedEntry). That would tell the caller whether something was overwritten or not which is why it returns a bool now.

Then DelItem and PopItem could both call putItem.

runtime/dict.go
```diff
@@ -624,6 +624,34 @@ func dictKeys(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return toDictUnsafe(args[0]).Keys(f).ToObject(), nil
 }
 
+func dictPop(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
```
trotterdylan: It'd be nice for *Dict to have a method for popping keys since that functionality can't be replicated atomically from the existing interface. So maybe implement *Dict.PopItem() and have dictPop call `d.PopItem(...)`.

aisk:
Ready for review.

runtime/set.go
```diff
@@ -81,7 +81,15 @@ func toSetUnsafe(o *Object) *Set {
 
 // Add inserts key into s. If key already exists then does nothing.
 func (s *Set) Add(f *Frame, key *Object) (bool, *BaseException) {
-	return s.dict.putItem(f, key, None)
+	origin, raised := s.dict.putItem(f, key, None)
+	if raised != nil {
+		return false, raised
+	}
+	added := false
```
trotterdylan: `return origin == nil, nil`

runtime/dict.go
```diff
@@ -326,25 +326,15 @@ func (d *Dict) incVersion() {
...
-			d.table.incUsed(-1)
-			d.incVersion()
-			deleted = true
-		}
+	if originValue != nil {
+		deleted = true
 	}
-	d.mutex.Unlock(f)
-	return deleted, raised
+	return deleted, nil
```
trotterdylan: You can get rid of the deleted local var: `return originValue != nil, nil`

runtime/dict.go
```diff
@@ -396,38 +392,49 @@ func (d *Dict) Len() int {
...
 		if v != d.version {
 			// Dictionary was recursively modified. Blow up instead
 			// of trying to recover.
 			raised = f.RaiseType(RuntimeErrorType, "dictionary changed during write")
 		} else {
-			if newTable, ok := t.writeEntry(f, index, &dictEntry{hash.Value(), key, value}); ok {
-				if newTable != nil {
-					d.storeTable(newTable)
+			if value == nil {
+				// Going to delete the entry
```
trotterdylan: Use full sentences including period at the end.

runtime/dict.go
```diff
@@ -396,38 +392,49 @@ func (d *Dict) Len() int {
 	return d.loadTable().loadUsed()
 }
 
-// putItem associates value with key in d, returning true if the key was added
-// (i.e. it was not already present in d).
-func (d *Dict) putItem(f *Frame, key, value *Object) (bool, *BaseException) {
+// putItem associates value with key in d, returning the old associalted value if
+// the key was added, or nil if it was not already present in d.
```
trotterdylan: "... returning the old *associated* value if ..."

aisk:
Updated.

PR #239, opened 2017-02-07
[Runtime] Implement str.isalpha(), str.isdigit(), str.isalnum()

corona10:
Implement str.isalpha(), str.isdigit(), str.isalnum().


runtime/str.go
```diff
@@ -918,6 +933,57 @@ func strInterpolate(f *Frame, format string, values *Tuple) (*Object, *BaseExcep
...
+	return GetBool(true).ToObject(), nil
+}
+
+func strIsDigit(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
+	if raised := checkMethodArgs(f, "isdigit", args, StrType); raised != nil {
+		return nil, raised
+	}
+	s := toStrUnsafe(args[0]).Value()
+	if len(s) == 0 {
+		return GetBool(false).ToObject(), nil
```
aisk: I think it's more suitable for just using the global bool instance `True` / `False.`

corona10: @aisk Ah awesome!! Thank you for your comment!!

runtime/str.go
```diff
@@ -136,6 +136,18 @@ func hashString(s string) int {
 	return h
 }
 
+func isAlNum(c byte) bool {
```
trotterdylan: Please group helpers at the bottom of the file.

runtime/str_test.go
```diff
@@ -347,6 +347,18 @@ func TestStrMethods(t *testing.T) {
 		{"find", wrapArgs("bar", "a", 0, -1), NewInt(1).ToObject(), nil},
 		{"find", wrapArgs("foo", newTestTuple("barfoo", "oo").ToObject()), nil, mustCreateException(TypeErrorType, "'find/index' requires a 'str' object but received a 'tuple'")},
 		{"find", wrapArgs("foo", 123), nil, mustCreateException(TypeErrorType, "'find/index' requires a 'str' object but received a 'int'")},
+		{"isalnum", wrapArgs("123abc"), True.ToObject(), nil},
+		{"isalnum", wrapArgs(""), False.ToObject(), nil},
+		{"isalnum", wrapArgs("#$%"), False.ToObject(), nil},
```
trotterdylan: Also include a test case where there some some characters are in alnum but others aren't, e.g. "abc#123". Same idea for isalpha and isdigit.

corona10:
@trotterdylan 
Thanks for reviews!!
I updated it

PTAL

PR #240, opened 2017-02-08
Add __builtin__.round

aisk:


runtime/builtin_types.go
```diff
@@ -827,3 +861,18 @@ func zipLongest(f *Frame, args Args) ([][]*Object, *BaseException) {
 	}
 	return result, nil
 }
+
+func round(number float64, places int) float64 {
+	var round float64
```
trotterdylan: This seems kind of crazy but then I read what CPython does and it's even more wild: https://github.com/python/cpython/blob/2.7/Objects/floatobject.c#L1092

My concern is that there will be subtle differences in behavior here with respect to CPython and it will only surface as part of divergent behavior in a complex program that will be difficult to debug. I would recommend either:

1. Write a dead-simple implementation for the purposes of unblocking unittest (e.g. Round() from [this gist comment](https://gist.github.com/DavidVaini/10308388#gistcomment-1391788))

2. Write a comprehensive implementation based on what's in floatobject.c linked above.

The PR as written is neither simple nor comprehensive so it's liable to end up causing us debugging headaches down the road.

runtime/builtin_types.go
```diff
@@ -570,6 +571,38 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+	var raised *BaseException
+	if argc == 1 {
+		expectedTypes = expectedTypes[:1]
+	}
+	if raised = checkFunctionArgs(f, "round", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	ndigits := 0
+	if argc > 1 {
+		ndigits, raised = IndexInt(f, args[1])
```
aisk: To my surprise, the CPython's `round` function do accept an object with `__index__` method as second parameter. Added a test case for this.

runtime/builtin_types.go
aisk: Just implemented it as `math.Floor(f + 0.5)`, and raises NotImplementError while passing second paramter on it. How about it?

aisk:
Updated, ready for review.

runtime/type_test.go
```diff
@@ -548,3 +548,12 @@ func newTestClassStrictEq(name string, base *Type) *Type {
 	}))
 	return t
 }
+
+func newTestIndexObject(index int) *Object {
```
trotterdylan: Not sure this belongs in type_test.go. Maybe just put it in builtin_types_test.go for now.

runtime/builtin_types.go
```diff
@@ -570,6 +571,34 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+		if raised != nil {
+			return nil, raised
+		}
+	}
+	// TODO: implement this.
+	if ndigits != 0 {
+		return nil, f.RaiseType(NotImplementedType, "round with ndigits is not implemented in grumpy")
```
trotterdylan: NotImplementedErrorType

runtime/builtin_types_test.go
```diff
@@ -275,6 +270,12 @@ func TestBuiltinFuncs(t *testing.T) {
 		{f: "repr", args: wrapArgs(NewUnicode("abc")), want: NewStr("u'abc'").ToObject()},
 		{f: "repr", args: wrapArgs(newTestTuple("foo", "bar")), want: NewStr("('foo', 'bar')").ToObject()},
 		{f: "repr", args: wrapArgs("a", "b", "c"), wantExc: mustCreateException(TypeErrorType, "'repr' requires 1 arguments")},
+		{f: "round", args: wrapArgs(1234.567), want: NewFloat(1235).ToObject()},
+		{f: "round", args: wrapArgs(1234.111), want: NewFloat(1234).ToObject()},
+		{f: "round", args: wrapArgs(-1234.567), want: NewFloat(-1235).ToObject()},
+		{f: "round", args: wrapArgs(-1234.111), want: NewFloat(-1234).ToObject()},
+		{f: "round", args: wrapArgs(1234.567, newTestIndexObject(0)), want: NewFloat(1235).ToObject()},
+		{f: "round", args: wrapArgs("foo"), wantExc: mustCreateException(TypeErrorType, "a float is required")},
```
trotterdylan: Include a test for the unimplemented ndigits > 0 case.

runtime/builtin_types.go
```diff
@@ -570,6 +571,34 @@ func builtinRepr(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
 	return s.ToObject(), nil
 }
 
+func builtinRound(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	argc := len(args)
+	expectedTypes := []*Type{ObjectType, ObjectType}
+	var raised *BaseException
```
trotterdylan: No need to declare this I don't think.

aisk:
Updated.

PR #241, opened 2017-02-08
[WIP]Add colorsys module

aisk:
Looks like unittest module is using `round` function interally, so please review #240 first.


aisk:
Already implemented, closed.

PR #243, opened 2017-02-09
Tag `attr_mode:"rw"` to read-write native attributes

alanjds:
Use tag `attr_mode:"rw"` to allow native attributes to be changed by Python code. The attribute should be public on Golang (start with a Capital).

- ``makeStructFieldDescriptor`` needs a last boolean argument. Please provide ``false`` for private fields.
- When the new argument is ``true`` and the struct field is public, it can be changed (set) via Python
- Will not be locked on changes. If locking is needed, do it in Python or via a custom Property setter


alanjds:
I signed it!

runtime/descriptor.go
```diff
@@ -98,11 +98,13 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
 
 // makeStructFieldDescriptor creates a descriptor with a getter that returns
 // the field given by fieldName from t's basis structure.
-func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object {
+func makeStructFieldDescriptor(t *Type, fieldName, propertyName string, writtable bool) *Object {
```
trotterdylan: I would suggest a new `type fieldDescriptorType int` along with fieldDescriptorRO and fieldDescriptorRW enums rather than a writeable bool. It makes calling code less mysterious.

runtime/descriptor.go
```diff
@@ -98,11 +98,13 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
 	field, ok := t.basis.FieldByName(fieldName)
 	if !ok {
 		logFatal(fmt.Sprintf("no such field %q for basis %s", fieldName, nativeTypeName(t.basis)))
 	}
+	nativeFieldType := getNativeType(field.Type)
```
trotterdylan: This is only used in one spot down below so move it down there.

runtime/descriptor.go
```diff
@@ -117,5 +119,31 @@ func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object
 		}
 		return ret, raised
 	}
+	if writtable {
+		setterFunc := func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+			ret := None
+			var raised *BaseException
+			if raised = checkFunctionArgs(f, fieldName, args, ObjectType, ObjectType); raised == nil {
```
trotterdylan: Prefer to [return early](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) rather than having nested if statements:

```
if raised := checkFunctionArgs(...); raised != nil {
  return nil, raised
}
```

runtime/descriptor.go
```diff
@@ -117,5 +119,31 @@ func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object
...
+				self := args[0]
+				newvalue := args[1]
+				if !newvalue.isInstance(nativeFieldType) {
```
trotterdylan: Could this be done as part of checkFunctionArgs above?

runtime/descriptor.go
```diff
@@ -117,5 +119,31 @@ func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object
...
+					val := t.slots.Basis.Fn(self).FieldByIndex(field.Index)
+
+					if converted, raised = maybeConvertValue(f, newvalue, val.Type()); raised == nil {
+						val.Set(converted)
+					}
+				}
+			}
+			return ret, raised
+		}
+		return newProperty(
```
trotterdylan: I would declare the setter above and set it inside the writeable if block:

```
setter := None
if writeable {
  ...
  setter = newBuiltinFunction("_set"+fieldName, setterFunc).ToObject()
}
return newProperty(newBuiltinFunction("_get"+fieldName, getterFunc).ToObject(), setter, None).ToObject()
```

runtime/descriptor.go
alanjds: You mean, an ``int`` from an ``enum`` instead of the ``writtable bool``? Well, this could accommodate more modes in the future if needed. Just confirm please.

trotterdylan: Right. It could accommodate more modes. The bigger advantage IMO is it makes the calling code more readable.

runtime/descriptor.go
alanjds: Sure.

runtime/descriptor.go
alanjds: I had followed the style of ``getterFunc()`` above, but can change mine or both to what you are asking. Please confirm.

trotterdylan: Ah. Thanks for being consistent. I'm not sure why that code is written that way. It'd be great to clean it up as well. Thanks!

runtime/descriptor.go
alanjds: I guess so, but may be less readable. Anyway I had followed the ``getterFunc()`` style. I could change mine or both. Please confirm.

runtime/descriptor.go
alanjds: Sure. I though that too, but was not sure if I need to type the ``setter := None``, nor that ``newBuiltinFunction`` returns an Object at the time this was coded.

alanjds: Ok... FunctionType extends ObjectType.

runtime/descriptor.go
alanjds: Eh... ok. Should I change (1) nothing, (2) mine, (3) everything? ;)

runtime/descriptor.go
trotterdylan: Right, you need to call ToObject on the result

trotterdylan: Right, you need to call ToObject on the result

runtime/descriptor.go
trotterdylan: Unless it proves to be really ugly, it'd be good to use early returns in your code. If you're feeling adventurous then also change the style of getterFunc.

runtime/descriptor.go
alanjds: Done on aa1e9f602c3737edf832cab6e8089311484122bc

runtime/descriptor.go
alanjds: Tried. Got too long line to be read for me

runtime/descriptor.go
alanjds: Done on a040008c45968849b52c83ff85083a0ccdb0a172

runtime/descriptor.go
alanjds: Done on b88c56917c317f069ba5a55b61446857c0ad3de4

runtime/descriptor.go
alanjds: Not done yet

alanjds:
There are tests failing because of an updated TypeError message. Will update later and fix the tests, ok?

Later = Before ready to merge, but not right now

runtime/descriptor.go
```diff
@@ -98,24 +98,62 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+		}
+
+		val.Set(converted)
+		return ret, raised
+	}
+
+	getter := newBuiltinFunction("_get"+fieldName, getterFunc).ToObject()
+	setter := None
+	if writtable {
+		setter = newBuiltinFunction("_set"+fieldName, setterFunc).ToObject()
```
trotterdylan: I would declare setterFunc in this block. It does make it a bit more indented than would be ideal, but it makes it clear that it's not used outside of here.

runtime/descriptor.go
```diff
@@ -98,24 +98,62 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
 		}
+
+		self := args[0]
+		if !self.isInstance(t) {
+			format := "descriptor '%s' for '%s' objects doesn't apply to '%s' objects"
+			raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), self.typ.Name()))
+			return ret, raised
+		}
+
+		ret, raised = WrapNative(f, t.slots.Basis.Fn(self).FieldByIndex(field.Index))
```
alanjds: Yes. Is this more readable?
Anyway, will be done.

runtime/descriptor.go
```diff
@@ -98,24 +98,62 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
 		return ret, raised
 	}
-	return newProperty(newBuiltinFunction("_get"+fieldName, getterFunc).ToObject(), None, None).ToObject()
+
+	setterFunc := func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+		ret := None
+		var raised *BaseException
+		if raised = checkFunctionArgs(f, fieldName, args, ObjectType, ObjectType); raised != nil {
+			return ret, raised
```
alanjds: Will the caller handle the ``nil, raised`` instead of the expected ``None, raised``?

I am trying to keep this compatible with the existing version, so am holding myself from external-facing changes.

Can be done. Please confirm.

runtime/descriptor.go
alanjds: But on https://github.com/google/grumpy/pull/243#discussion_r100479160 you asked me to pull it out from the block. Or had I misunderstood something?

Can be done. Please confirm.

runtime/descriptor.go
trotterdylan: I think it's more readable.

runtime/descriptor.go
trotterdylan: If raised is non-nil, then the first return value should be nil. It protects against callers that assume raised is nil if the first return value is non-nil. In that case they could fail to re-raise or to call ExcRestore().

runtime/descriptor.go
trotterdylan: In that case I was trying to avoid multiple returns that were very similar:

```
  return newProperty(...)
}
return newProperty(...)
```

Where `...` had a lot of repetition. I think what you have now is an improvement, I'm just suggesting moving the setterFunc definition inside the if block.

alanjds: Done on 365d7a7

runtime/descriptor.go
alanjds: And if it does not except, should ``return ret, nil`` instead of ``return ret, raised``?

runtime/descriptor.go
alanjds: Done in 678631a

runtime/descriptor.go
alanjds: Done in 678631a

runtime/descriptor.go
alanjds: Done in 7019392

alanjds:
@trotterdylan I guess that there is only the TypeError tests yet to be fixed now.

runtime/descriptor.go
```diff
@@ -19,6 +19,13 @@ import (
 	"reflect"
 )
 
+type fieldDescriptorType int
+
+const (
+    FIELDDESCRIPTOR_RO fieldDescriptorType = iota
```
trotterdylan: Const names should use the same naming convention as variable names. Also, these should be private to the package, since makeStructFieldDescriptor is not public anyway: fieldDescriptorRO, fieldDescriptorRW

runtime/descriptor.go
```diff
@@ -98,24 +105,60 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
 // makeStructFieldDescriptor creates a descriptor with a getter that returns
 // the field given by fieldName from t's basis structure.
-func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object {
+func makeStructFieldDescriptor(t *Type, fieldName, propertyName string, fieldtype fieldDescriptorType) *Object {
 	field, ok := t.basis.FieldByName(fieldName)
 	if !ok {
 		logFatal(fmt.Sprintf("no such field %q for basis %s", fieldName, nativeTypeName(t.basis)))
 	}
+
+	getter := None
```
trotterdylan: Declare this down below where it gets used.

runtime/descriptor.go
```diff
@@ -98,24 +105,60 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
-			o := args[0]
-			if !o.isInstance(t) {
+		if raised = checkFunctionArgs(f, fieldName, args, ObjectType); raised != nil {
+			return nil, raised
+		}
+
+		self := args[0]
+		if !self.isInstance(t) {
+			format := "descriptor '%s' for '%s' objects doesn't apply to '%s' objects"
+			raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), self.typ.Name()))
```
trotterdylan: Don't bother assigning to raised, just return the value:

```
return nil, f.RaiseType(...)
```

runtime/descriptor.go
```diff
@@ -98,24 +105,60 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+			return nil, raised
+		}
+
+		return WrapNative(f, t.slots.Basis.Fn(self).FieldByIndex(field.Index))
+	}
+
+	if fieldtype == FIELDDESCRIPTOR_RW {
+		setterFunc := func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+			var raised *BaseException
```
trotterdylan: Don't bother declaring this, just assign to it as needed:

```
if raised := checkFunctionArgs(...) {
```

runtime/descriptor.go
```diff
@@ -98,24 +105,60 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+			}
+
+			self := args[0]
+			newvalue := args[1]
+			if !newvalue.isInstance(getNativeType(field.Type)) {
 				format := "descriptor '%s' for '%s' objects doesn't apply to '%s' objects"
-				raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), o.typ.Name()))
-			} else {
-				ret, raised = WrapNative(f, t.slots.Basis.Fn(o).FieldByIndex(field.Index))
+				raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), newvalue.typ.Name()))
```
trotterdylan: Don't assign, just return

runtime/descriptor.go
```diff
@@ -98,24 +105,60 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
 
 // makeStructFieldDescriptor creates a descriptor with a getter that returns
 // the field given by fieldName from t's basis structure.
-func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object {
+func makeStructFieldDescriptor(t *Type, fieldName, propertyName string, fieldtype fieldDescriptorType) *Object {
```
trotterdylan: fieldType

runtime/type.go
```diff
@@ -153,7 +153,11 @@ func prepareBuiltinType(typ *Type, init builtinTypeInit) {
 		for i := 0; i < numFields; i++ {
 			field := basis.Field(i)
 			if attr := field.Tag.Get("attr"); attr != "" {
-				dict[attr] = makeStructFieldDescriptor(typ, field.Name, attr)
+				writtable := FIELDDESCRIPTOR_RO
```
trotterdylan:
```
fieldType := fieldDescriptorRO
```

runtime/descriptor.go
alanjds: Really? This is so ugly and impossible to distinguish constants from variables. Ok... done on 845490d91

runtime/descriptor.go
alanjds: Done on 845490d91

runtime/descriptor.go
alanjds: Done on 66740c4

runtime/descriptor.go
alanjds: Done on 15c9962

runtime/descriptor.go
alanjds: Done on 15c9962. Except on ``if converted, raised = maybeConvertValue()...`` because I need the ``converted`` later.

runtime/descriptor.go
alanjds: Done on 15c9962

runtime/type.go
alanjds: Done on 845490d91

runtime/descriptor.go
trotterdylan: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

runtime/descriptor.go
```diff
@@ -98,24 +105,54 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
 
 // makeStructFieldDescriptor creates a descriptor with a getter that returns
 // the field given by fieldName from t's basis structure.
-func makeStructFieldDescriptor(t *Type, fieldName, propertyName string) *Object {
+func makeStructFieldDescriptor(t *Type, fieldName, propertyName string, fieldType fieldDescriptorType) *Object {
```
trotterdylan: When fieldType is fieldDesciptorRW, you should check that `field.PkgPath` is empty (i.e. the [field is exported](https://golang.org/pkg/reflect/#StructField) and therefore settable). Otherwise we should logFatal. This will allow us to avoid panicking at run time when we accidentally made an unexported field RW in this library.

runtime/descriptor.go
```diff
@@ -98,24 +105,54 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+			}
+
+			self := args[0]
+			newvalue := args[1]
+			if !newvalue.isInstance(getNativeType(field.Type)) {
 				format := "descriptor '%s' for '%s' objects doesn't apply to '%s' objects"
-				raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), o.typ.Name()))
-			} else {
-				ret, raised = WrapNative(f, t.slots.Basis.Fn(o).FieldByIndex(field.Index))
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), newvalue.typ.Name()))
```
trotterdylan: Shouldn't this be self.typ.Name(), not newvalue.typ.Name()?

runtime/descriptor.go
alanjds: I though about testing against ``.canSet()`` as seen on http://stackoverflow.com/a/6396678/798575. Is it better to check ``field.PkgPath`` instead? Please confirm

runtime/descriptor.go
```diff
@@ -98,24 +105,54 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+
+	setter := None
+	if fieldType == fieldDescriptorRW {
+		setterFunc := func(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+			if raised := checkFunctionArgs(f, fieldName, args, ObjectType, ObjectType); raised != nil {
+				return nil, raised
+			}
+
+			self := args[0]
+			newvalue := args[1]
```
alanjds: This was wrote as one word for me, not two. So, not newValue instead of new_value. Just newvalue. Anyway, will be done.

runtime/descriptor.go
trotterdylan: AFAICT canSet is not available on the field, only the value. Since the value is only available in the `__set__` call, not during creation of the descriptor, the best we could do is a runtime check and raise an exception. That's better than panicking, but I think it's better to logFatal during Grumpy package initialization so that we can detect the error as early as possible.

runtime/descriptor.go
alanjds: I really do not know. This is from the past version, just renamed the newvalue. Will update to ``self``

runtime/descriptor.go
alanjds: Ok. Will be done. Thanks for pointing.

runtime/descriptor.go
alanjds: ok. Will be done.

runtime/descriptor.go
alanjds: Done on d67d995

runtime/descriptor.go
alanjds: Done on 047d270

runtime/descriptor.go
alanjds: Done on 062669c

alanjds:
Finally. Existing tests passes and gofmt smiles.

Now, write new tests.

Question: is there a coverage tool being used on this project? Or should I start [with this](http://stackoverflow.com/questions/10516662/how-to-measure-code-coverage-in-golang)?

trotterdylan:
`make cover` should tell you which new or changed lines lack coverage.

alanjds:
Hum... told not.

```
$ make cover
ok  	grumpy	1.273s	coverage: 92.6% of statements
$
```

trotterdylan:
Ah, sorry, I forgot that it will calculate changed lines using `git diff HEAD` by default, which in your case is probably no diffs. You need to specify a baseline commit against which it can calculate changed lines, like `make DIFF_COMMIT=83bb953 cover` (83bb953 is HEAD for the google/grumpy repo, you may need to pick a different baseline commit from your own history).

trotterdylan:
Or if you want an exhaustive list of uncovered lines in the grumpy lib (this will include lines you haven't changed), you can run:

```
PKG_DIR=build/pkg/$(go env GOOS GOARCH | sed 'N;s/\n/_/')
make $PKG_DIR/grumpy.cover
tools/coverparse $PKG_DIR/grumpy.cover
```

alanjds:
Ah ok. Thanks.

Now I have some work to do:

```
$ make DIFF_COMMIT=43ddec8deffd cover
runtime/descriptor.go:115: missing coverage
runtime/descriptor.go:116: missing coverage
runtime/descriptor.go:117: missing coverage
runtime/descriptor.go:129: missing coverage
runtime/descriptor.go:130: missing coverage
runtime/descriptor.go:131: missing coverage
runtime/descriptor.go:132: missing coverage
runtime/descriptor.go:134: missing coverage
runtime/descriptor.go:135: missing coverage
runtime/descriptor.go:136: missing coverage
runtime/descriptor.go:137: missing coverage
runtime/descriptor.go:139: missing coverage
runtime/descriptor.go:140: missing coverage
runtime/descriptor.go:141: missing coverage
runtime/descriptor.go:143: missing coverage
runtime/descriptor.go:146: missing coverage
runtime/descriptor.go:147: missing coverage
runtime/descriptor.go:148: missing coverage
runtime/descriptor.go:149: missing coverage
runtime/descriptor.go:150: missing coverage
runtime/descriptor.go:151: missing coverage
runtime/descriptor.go:153: missing coverage
runtime/descriptor.go:154: missing coverage
runtime/descriptor.go:157: missing coverage
runtime/type.go:157: missing coverage
runtime/type.go:158: missing coverage
runtime/type.go:159: missing coverage
make: *** [cover] Error 1
$
```

alanjds:
Got tests, but I do not know how to cover the remain lines. Tried to call something intentionally wrong, but had not hit the lines desired, after struggling with the existing test structure. Any ideas to increase coverage? 

Is this enough for accepting the PR?

```
$ make DIFF_COMMIT=43ddec8deffd cover
ok  	grumpy	1.265s	coverage: 92.9% of statements
runtime/descriptor.go:115: missing coverage
runtime/descriptor.go:116: missing coverage
runtime/descriptor.go:117: missing coverage
runtime/descriptor.go:130: missing coverage
runtime/descriptor.go:131: missing coverage
runtime/descriptor.go:132: missing coverage
runtime/descriptor.go:135: missing coverage
runtime/descriptor.go:136: missing coverage
runtime/descriptor.go:137: missing coverage
runtime/descriptor.go:157: missing coverage
runtime/descriptor.go:158: missing coverage
runtime/descriptor.go:159: missing coverage
make: *** [cover] Error 1
$
```

trotterdylan:
```
runtime/descriptor.go:115: missing coverage
runtime/descriptor.go:116: missing coverage
runtime/descriptor.go:117: missing coverage
```

It's fine to leave these untested. The reason it's hard to trigger these lines is that the property always invokes the function the correct way. You'd need to fetch the getter function and then call it incorrectly to trigger this code path. Something like this:

```
>>> class A(object):
...  @property
...  def p(self):
...   return 3
... 
>>> A.p.fget(1, 2, 3)
TypeError: p() takes exactly 1 argument (3 given)
```

But doing all this probably adds more complexity without a lot of value.

```
runtime/descriptor.go:130: missing coverage
runtime/descriptor.go:131: missing coverage
runtime/descriptor.go:132: missing coverage
```

Don't worry about these either. This condition is basically an assertion and if it triggers then we've got bigger issues than the logFatal not functioning properly.

```
runtime/descriptor.go:135: missing coverage
runtime/descriptor.go:136: missing coverage
runtime/descriptor.go:137: missing coverage
```

This lines are basically the same as the first set above so ignore these.

```
runtime/descriptor.go:157: missing coverage
runtime/descriptor.go:158: missing coverage
runtime/descriptor.go:159: missing coverage
```

It should be possible to trigger this by attempting to set the property to Go type that isn't convertible to the underlying field type. If the field is an int, try to set it with a string value, for example.

alanjds:
Right. Will cover 157-159.

Will try a Python test, like the ones on the `testing/` folder. Just set an str to ``file.softspace`` should do...

alanjds:
Not increased? Had I misunderstood what you said?

```
157  if converted, raised = maybeConvertValue(f, newValue, val.Type()); raised != nil {
158  	return nil, raised
159  }
```

> It should be possible to trigger this by attempting to set the property to Go type that isn't convertible to the underlying field type. If the field is an int, try to set it with a string value, for example.

I tried to test via: ``testing/file_test.py``
```
try:
    f.softspace = 'deadbeef'
except TypeError as e:
    assert str(e) == "an int is required. Field 'file.softspace' cannot store a 'str' value"
```

trotterdylan:
Coverage is only calculated from tests within the runtime directory. Not from tests in generated code. It'd be nice to get this to work but I recall it being a bit of pain to get working.

You should be able to trigger this by doing something like: `SetAttr(f, o, NewStr("softspace"), NewStr("deadbeef").ToObject())`

alanjds:
> ...but I recall it being a bit of pain to get working
Oh. But I guess it to be worth, as I see to be very easier to test stuff via Python. I feel it being very harder to beat the roadmap of July 2017 without it. Maybe this could be my next ticket. Can you discourse why is hard/painful to do it and make me give up please?

trotterdylan:
Just the way `go test` works, it expects `*_test.go` files with the package name of that being tested. And then within those files it expects Test* functions. One way to do it would be to generate a module_test.go file with a single TestModule function in there that loads the module and executes it as main. So I guess it's not that hard, it's just a bit of work.

alanjds:
I took out the existing descriptor tests file and ``rm -fr build``.
Ran the cover: 33 new lines uncovered

```sh
$ make DIFF_COMMIT=43ddec8deffd cover | wc -l
33
```
Ran the ``make test``. I wrote a ``file_test.go`` by hand and put aside ``file.go`` on the ``build/testing`` folder.

```golang
# build/testing/file_test.go
package grumpy

import (
    "testing"
)

func TestFile(t *testing.T) {
    main()
}
```

Ran the cover. Same 33 lines missing.
```
$ make DIFF_COMMIT=43ddec8deffd cover | wc -l
33
```

Are the ``cover`` ignoring the ``build/testing`` folder?

alanjds:
Gave up on cover Python for now. Now, back to the coverage:

```golang
157  if converted, raised = maybeConvertValue(f, newValue, val.Type()); raised != nil {
158  	return nil, raised
159  }
```
> It should be possible to trigger this by attempting to set the property to Go type that isn't convertible to the underlying field type. If the field is an int, try to set it with a string value, for example.

I do not know where to put the code you suggest, but I guess this is already being tested by the one in line 156.

```golang
// descriptor_test.go
156		{args: wrapArgs(FileType, "Softspace", "softspace", newObject(FileType), "wrong"), wantExc: mustCreateException(TypeErrorType, "an int is required. Field 'file.softspace' cannot store a 'str' value")},
```

Is this [code up] not equivalent to your suggestion [code down]? (your "deadbeef" = mine "wrong")

> ``SetAttr(f, o, NewStr("softspace"), NewStr("deadbeef").ToObject())``

alanjds:
Travis build output:
```
compiler/imputil_test PASS
compiler/util_test PASS
testing/assert_test_py PASS
testing/assign_test_py PASS
testing/builtin_test_py PASS
testing/class_test_py PASS
testing/compare_test_py PASS
testing/complex_test_py PASS
testing/comprehension_test_py PASS
testing/dict_test_py PASS
  File "testing/file_test.py", line 10, in <module>
    assert str(e) == "an int is required. Field 'file.softspace' cannot store a 'str' value", e
AssertionError: an integer is required
make: *** [build/testing/file_test_py.pass] Error 1
make: *** Waiting for unfinished jobs....
The command "make gofmt lint && make -j2 test" exited with 2.
Done. Your build exited with 1.
```

This is strange. Is Travis testing using CPython instead of Grumpy?

trotterdylan:
Yes, all the tests in testing are run under CPython and Grumpy to ensure they behave the same way. The targets ending in \_py are run under CPython.

trotterdylan:
Right, I believe line 156 that you mentioned should test that case. Is it showing as covered?

alanjds:
No. As not covered.

I believe that this is not testable as is not a possible path for valid
Python code. May be a possible path via Go code I guess, but I was not able
to bake it.

Il 22 apr 2017 5:07 PM, "Dylan Trotter" <notifications@github.com> ha
scritto:

runtime/type.go
```diff
@@ -153,7 +153,11 @@ func prepareBuiltinType(typ *Type, init builtinTypeInit) {
 		for i := 0; i < numFields; i++ {
 			field := basis.Field(i)
 			if attr := field.Tag.Get("attr"); attr != "" {
-				dict[attr] = makeStructFieldDescriptor(typ, field.Name, attr)
+				writtable := fieldDescriptorRO
```
trotterdylan: writeable

trotterdylan: Maybe call this variable descType or something like that.

runtime/descriptor_test.go
```diff
@@ -110,3 +110,57 @@ func TestMakeStructFieldDescriptor(t *testing.T) {
...
+			return nil, raised
+		}
+		return set.Call(f, wrapArgs(args[3], args[4]), nil)
+	}).ToObject()
+	cases := []invokeTestCase{
+		{args: wrapArgs(FileType, "Softspace", "softspace", newObject(FileType), NewInt(0).ToObject()), want: None},
+		{args: wrapArgs(FileType, "Softspace", "softspace", newObject(FileType), NewInt(0)), want: None},
+		{args: wrapArgs(FileType, "Softspace", "softspace", newObject(FileType), "wrong"), wantExc: mustCreateException(TypeErrorType, "an int is required. Field 'file.softspace' cannot store a 'str' value")},
+		{args: wrapArgs(FileType, "Softspace", "softspace", 42, NewInt(0)), wantExc: mustCreateException(TypeErrorType, "descriptor 'softspace' for 'file' objects doesn't apply to 'int' objects")},
+		//{args: wrapArgs(StrType, "value", "value", 42), wantExc: mustCreateException(TypeErrorType, "field 'dict' is not public on Golang code. Please fix it.")},
```
trotterdylan: Remove commented out code.

runtime/descriptor_test.go
```diff
@@ -110,3 +110,57 @@ func TestMakeStructFieldDescriptor(t *testing.T) {
 		}
 	}
 }
+
+func TestMakeStructFieldDescriptorRW_get(t *testing.T) {
```
trotterdylan: TestMakeStructFieldDescriptorRWGet

runtime/descriptor.go
```diff
@@ -98,24 +105,66 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+
+			val.Set(converted)
+			return None, nil
 		}
-		return ret, raised
+
+		setter = newBuiltinFunction("_set"+fieldName, setterFunc).ToObject()
 	}
-	return newProperty(newBuiltinFunction("_get"+fieldName, getterFunc).ToObject(), None, None).ToObject()
+	getter := newBuiltinFunction("_get"+fieldName, getterFunc).ToObject()
```
trotterdylan: Define getter right below getterFunc

runtime/descriptor.go
```diff
@@ -98,24 +105,66 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
+
+			if !self.isInstance(t) {
 				format := "descriptor '%s' for '%s' objects doesn't apply to '%s' objects"
-				raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), o.typ.Name()))
-			} else {
-				ret, raised = WrapNative(f, t.slots.Basis.Fn(o).FieldByIndex(field.Index))
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), self.typ.Name()))
+			}
+
+			if !newValue.isInstance(getNativeType(field.Type)) {
```
trotterdylan: Why is this necessary? This check should be handled implicitly by the maybeConvertValue() call below. As written, I believe that this will reject assignments between compatible types, like assigning to a field of `type Foo int` with an `int` value.

alanjds:
Just to notice: I will start spending time on having existing PRs merged instead of opening new ones, as soon as I get the time needed.

Starting by this one here.

trotterdylan:
Sounds good. Please ping the PRs when Jenkins is passing and they're ready to be reviewed. I've lost track of the status of some of the PRs.

alanjds:
Huh. The missing lines increased? Should I try to rebase over actual master?

runtime/type.go
alanjds: Done on 012c13770740

runtime/descriptor_test.go
alanjds: Done on 4c289860b5b1f

runtime/descriptor.go
alanjds: Done on 679e1d7

runtime/descriptor_test.go
alanjds: Done on 679e1d7

runtime/descriptor.go
alanjds: Done on 2121ac1

runtime/descriptor.go
alanjds: The CPython traceback setting file.softspace to a string is: ``TypeError: an integer is required``.

For now, I am raising ``"an %s is required. Field '%s.%s' cannot store a '%s' value"``, just to be more helpful.

```python
# testing/file_test.py
f = open('/tmp/foo', 'w')
assert f.softspace == 0

f.softspace = 1
assert f.softspace == 1

try:
    f.softspace = 'deadbeef'
except TypeError as e:
    if 'is required' not in str(e):
        raise e     # Wrong exception arrived to us!
else:
    raise RuntimeError('a TypeError should had raised.')
```

Without this pre-check, Grumpy raises an ``TypeError: cannot convert string to int`` instead. 

And more intriguing: changing the sample code to ``f.softspace = '1'``, Grumpy still gives ``TypeError: cannot convert string to int``. CPython keeps on ``TypeError: an integer is required``.

alanjds: I can move the ``!newValue.isInstance`` check into the ``maybeConvertValue`` "except" block, to save a check on the happy path I guess...

Or change the maybeConvertValue error message instead maybe.

alanjds:
```bash
$ make DIFF_COMMIT=d7fd9ac cover
ok  	grumpy	1.381s	coverage: 92.6% of statements
runtime/descriptor.go:115: missing coverage
runtime/descriptor.go:116: missing coverage
runtime/descriptor.go:117: missing coverage
runtime/descriptor.go:131: missing coverage
runtime/descriptor.go:132: missing coverage
runtime/descriptor.go:133: missing coverage
runtime/descriptor.go:136: missing coverage
runtime/descriptor.go:137: missing coverage
runtime/descriptor.go:138: missing coverage
make: *** [cover] Error 1
```

From what is said on https://github.com/google/grumpy/pull/243#issuecomment-291989230, this should be enough coverage.

@trotterdylan can you take a look please?

alanjds:
-_-
Oops. Some test is broken in the CI

alanjds:
Ok. I have no clue why this broke:

```
compiler/stmt_test 16of16 PASS
Exception: foo
--- FAIL: TestRunMain (0.00s)
	module_test.go:300: runMainRedirectStderr() failed: read |0: bad file descriptor
RuntimeError: foo
FAIL
FAIL	grumpy	1.194s
make: *** [build/runtime.pass] Error 1
make: *** Waiting for unfinished jobs....
```

alanjds:
Travis, you are insane.

@trotterdylan Ok. Now Travis is happy. You can take a look.

runtime/descriptor_test.go
```diff
@@ -110,3 +110,55 @@ func TestMakeStructFieldDescriptor(t *testing.T) {
...
+	}
+	for _, cas := range cases {
+		if err := runInvokeTestCase(fun, &cas); err != "" {
+			t.Error(err)
+		}
+	}
+}
+
+func TestMakeStructFieldDescriptorRWSet(t *testing.T) {
+	//e := mustNotRaise(RuntimeErrorType.Call(NewRootFrame(), wrapArgs("foo"), nil))
```
trotterdylan: Remove commented-out code.

runtime/descriptor_test.go
```diff
@@ -110,3 +110,55 @@ func TestMakeStructFieldDescriptor(t *testing.T) {
...
+		get, raised := GetAttr(f, desc, NewStr("__get__"), nil)
+		if raised != nil {
+			return nil, raised
+		}
+		return get.Call(f, wrapArgs(args[3], t), nil)
+	}).ToObject()
+	cases := []invokeTestCase{
+		{args: wrapArgs(FileType, "Softspace", "softspace", newObject(FileType)), want: NewInt(0).ToObject()},
+		{args: wrapArgs(FileType, "Softspace", "softspace", 42), wantExc: mustCreateException(TypeErrorType, "descriptor 'softspace' for 'file' objects doesn't apply to 'int' objects")},
+		//{args: wrapArgs(FileType, "open", "open", true), wantExc: mustCreateException(TypeErrorType, "field 'open' is not public on Golang code. Please fix it.")},
```
trotterdylan: Remove this. The reason this doesn't work is that logFatal() doesn't raise an exception, it aborts the program.

runtime/descriptor.go
```diff
@@ -98,24 +105,61 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
-				ret, raised = WrapNative(f, t.slots.Basis.Fn(o).FieldByIndex(field.Index))
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), self.typ.Name()))
 			}
+
+			var converted reflect.Value
+			var raised *BaseException
+			val := t.slots.Basis.Fn(self).FieldByIndex(field.Index)
+			if converted, raised = maybeConvertValue(f, newValue, field.Type); raised != nil {
+				format := "an %s is required. Field '%s.%s' cannot store a '%s' value"
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format,
```
trotterdylan: Just return the exception that was raised by maybeConvertValue() rather than raising a new exception.

runtime/descriptor.go
```diff
@@ -98,24 +105,61 @@ func propertySet(f *Frame, desc, inst, value *Object) *BaseException {
...
-				raised = f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), o.typ.Name()))
-			} else {
-				ret, raised = WrapNative(f, t.slots.Basis.Fn(o).FieldByIndex(field.Index))
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format, propertyName, t.Name(), self.typ.Name()))
 			}
+
+			var converted reflect.Value
+			var raised *BaseException
+			val := t.slots.Basis.Fn(self).FieldByIndex(field.Index)
+			if converted, raised = maybeConvertValue(f, newValue, field.Type); raised != nil {
```
trotterdylan: move `converted, raised := maybeConvertValue(...)` outside of the if so that you don't need to bother declaring converted and raised explicitly above.

runtime/descriptor_test.go
alanjds: Done on c680142

runtime/descriptor_test.go
alanjds: Done on c680142

runtime/descriptor.go
alanjds: Done on 1ae6ea3

runtime/descriptor.go
alanjds: This I should not do because what I explained on https://github.com/google/grumpy/pull/243#discussion_r119990905 and https://github.com/google/grumpy/pull/243#discussion_r119990963.

If I do what you say, the error message would be ``TypeError: cannot convert string to int`` instead of ``TypeError: an int is required...``. Will be very different from CPython and nonsensical for convertible strings like ``f.softspace = '4321'`` for example

alanjds: Should I change [maybeConvertValue error message](https://github.com/google/grumpy/blob/master/runtime/native.go#L504) to something like ``fmt.Sprintf("an %s is required. %s was provided", expectedRType, rtype))`` instead?

trotterdylan: I think improving maybeConvertValue()'s error message is a good idea.

I don't think that extra context is very important though, since the traceback should provide enough info to determine the problem. CPython does not provide much information in the error message but it's still obvious what's wrong:

```
>>> file.__new__(file).softspace = 'foo'
TypeError: an integer is required
```

alanjds: Changed the error message on 812145c and done on c95237c.

alanjds:
@trotterdylan Travis is happy. Lets merge it!

PR #245, opened 2017-02-10
[Runtime] Implementation of str.isspace(), isupper(), islower(), istitle()

corona10:
Implementation of str.isspace(), str.isupper(), str.islower(), str.istitle().

For the `istitle()` operation, I adopted a CPython's implementation from [here](https://github.com/python/cpython/blob/2.7/Objects/stringobject.c#L3504).


runtime/str.go
```diff
@@ -1199,6 +1292,17 @@ func isDigit(c byte) bool {
 	return '0' <= c && c <= '9'
 }
 
+func isLower(c byte) bool {
+	return 'a' <= c && c <= 'z'
+}
+
+func isSpace(c byte) bool {
+	return c == ' ' || c == '\n' || c == '\t' || c == '\v' || c == '\f' || c == '\r'
```
trotterdylan: This may be more efficient as:

```
switch c {
  case ' ', '\n', '\t', '\v', '\f', '\r':
    return true
  default:
    return false
}
```

runtime/str.go
```diff
@@ -1199,6 +1292,17 @@ func isDigit(c byte) bool {
 	return '0' <= c && c <= '9'
 }
 
+func isLower(c byte) bool {
+	return 'a' <= c && c <= 'z'
```
trotterdylan: Can you replace the other identical comparisons in this file (e.g. in isAlpha and strTitle, toUpper) with a call to this function? Ditto for isUpper().

runtime/str.go
```diff
@@ -383,6 +383,95 @@ func strIsDigit(f *Frame, args Args, kwargs KWArgs) (*Object, *BaseException) {
...
+		return nil, raised
+	}
+
+	s := toStrUnsafe(args[0]).Value()
+	if len(s) == 0 {
+		return False.ToObject(), nil
+	}
+
+	if len(s) == 1 {
+		if isUpper(s[0]) {
```
trotterdylan: `return getBool(isUpper(s[0])).ToObject(), nil`

PR #247, opened 2017-02-13
Switch pyPrint() into based on *grumpy.File

corona10:
This PR is changing pyPirnt() into based on *grumpy.File.
It is related to PR #233 which is discussions about Stdout and Stdin.
I will proceed PR #233 after this PR is merged.

Related PR= #233

### DIFF
--- a/runtime/core.go
+++ b/runtime/core.go
@@ -16,7 +16,6 @@ package grumpy
 
 import (
 	"fmt"
-	"io"
 	"log"
 	"os"
 	"reflect"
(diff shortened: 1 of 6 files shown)


runtime/core.go
```diff
@@ -1216,17 +1215,17 @@ func hashNotImplemented(f *Frame, o *Object) (*Object, *BaseException) {
 }
 
 // pyPrint encapsulates the logic of the Python print function.
-func pyPrint(f *Frame, args Args, sep, end string, file io.Writer) *BaseException {
+func pyPrint(f *Frame, args Args, sep, end string, file *File) *BaseException {
 	for i, arg := range args {
 		if i > 0 {
-			fmt.Fprint(file, sep)
+			file.writeString(sep)
```
trotterdylan: Would you mind adding the code to raise IOErrorType for these calls? The previous code should have been checking this. It'd be good to take care of this while we're in here.

corona10:
@trotterdylan Updated!! PTAL. Thank's for comment.

PR #249, opened 2017-02-14
Switch os.Stderr to *grumpy.Stderr.

corona10:
After #247 is merged, we could change os.Stderr to *grumpy.Stderr.


corona10:
@trotterdylan This PR could be merged right now?

PR #253, opened 2017-02-16
Add unary plus

S-YOU:


runtime/core_test.go
```diff
@@ -795,6 +795,20 @@ func TestOct(t *testing.T) {
 	}
 }
 
+func TestPos(t *testing.T) {
+	cases := []invokeTestCase{
+		{args: wrapArgs(42), want: NewInt(42).ToObject()},
+		{args: wrapArgs(1.2), want: NewFloat(1.2).ToObject()},
+		{args: wrapArgs(NewLong(big.NewInt(123))), want: NewLong(big.NewInt(123)).ToObject()},
+		{args: wrapArgs("foo"), wantExc: mustCreateException(TypeErrorType, "bad operand type for unary +: 'str'")},
```
trotterdylan: Would you mind adding a test for a type that implements `__pos__`?

S-YOU: Added one, something like that?

PR #254, opened 2017-02-16
[WIP]Wildcard imports

claudiu-coman:
Hi @trotterdylan! I read the conversation in #192 and I thought I'd take a shot at wildcard imports. The code should be functional, I've updated the tests too.
If the idea is the right one, the next step for me would be to start trimming that code down. I can isolate some code in reusable methods and remove some of the duplication.

Before I take this PR further, can you tell me if I'm heading in the right direction?

### DIFF
--- a/compiler/stmt.py
+++ b/compiler/stmt.py
@@ -378,13 +378,11 @@ class StatementVisitor(algorithm.Visitor):
         self.block.bind_var(self.writer, asname, mod.expr)
 
   def visit_ImportFrom(self, node):
-    # Wildcard imports are not yet supported.
+    self._write_py_context(node.lineno)
     for alias in node.names:
       if alias.name == '*':
-        msg = 'wildcard member import is not implemented: from %s import %s' % (
-            node.module, alias.name)
-        raise util.ParseError(node, msg)
-    self._write_py_context(node.lineno)
+        self._import_wildcard(node)
+        return
     if node.module.startswith(_NATIVE_MODULE_PREFIX):
       values = [alias.name for alias in node.names]
       with self._import_native(node.module, values) as mod:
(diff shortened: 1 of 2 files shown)


compiler/stmt_test.py
```diff
@@ -366,14 +366,25 @@ def testImportFromFutureParseError(self):
...
+        from re import *
+        print compile"""))
+    self.assertEqual(0, result[0])
+    self.assertIn('<function compile at',result[1])
+
+    self.assertEqual((0, 'False\n'), _GrumpRun(textwrap.dedent("""\
+        from re import *
+        print 'sre_compile' in globals()""")))
+
+    # 'time' does not define __all__
```
trotterdylan: This seems like an implementation detail that could change. Maybe we need to create a couple dummy test modules for the purposes of testing this functionality.

compiler/stmt.py
```diff
@@ -748,6 +925,19 @@ def _import_native(self, name, values):
...
 
+  def _import_wildcard(self, node):
+    """If __all__ is defined, attempts to bind everything from it,
+    falls back on __dict__ otherwise.
+
+    Note: items in __dict__ whose names start with '_' are skipped.
+    """
+    module_name = node.module
+
+    with self.block.alloc_temp() as members, \
```
trotterdylan: IIUC members is a temporary that will be filled up, used to populate the globals dict and then thrown away. I would just populate the globals dict as I iterate over the imported module's members and avoid the temporary.

claudiu-coman:
It struck me that I could write the code in Go, as I got more familiar with the codebase. It was a bit too late and I was closer to finishing the Python code. And I was afraid Go would be the better solution. :) I'll start working on that on Monday, excited about how this will turn out.

BTW, the Travis jobs for my commits seem to take a lot of time. Job #679 seems to have been running for over 50 mins now. Is that normal?

trotterdylan:
Sounds good.

The OS X job on Travis is slow. Usually it suffices to validate that the
Linux job was successful. It's usually pretty quick.

PR #258, opened 2017-02-19
Update mktime to return local time result, not utc

trotterdylan:
Per the docs, mktime expects a time struct in local time, not UTC:
https://docs.python.org/2/library/time.html#time.mktime

This mismatch was causing TestDate.test_fromtimestamp in
test_datetime.py to fail on some systems.

### DIFF
--- a/lib/time.py
+++ b/lib/time.py
@@ -14,7 +14,7 @@
 
 """Time access and conversions."""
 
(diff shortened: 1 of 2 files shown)


trotterdylan:
@S-YOU Let me know if this change makes sense.

S-YOU:
Thank you @trotterdylan , I think I missed to check the docs.

PR #260, opened 2017-02-19
Implement __builtins__.sum

aisk:
Just implemented it as cpython dose: https://github.com/python/cpython/blob/2.7/Python/bltinmodule.c#L2307 .

### DIFF
--- a/runtime/builtin_types.go
+++ b/runtime/builtin_types.go
@@ -650,6 +650,85 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return result, nil
 }
 
+func builtinSum(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	argc := len(args)
+	expectedTypes := []*Type{ObjectType, ObjectType}
+	if argc == 1 {
+		expectedTypes = expectedTypes[:1]
+	}
+	if raised := checkFunctionArgs(f, "sum", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	iter, raised := Iter(f, args[0])
+	if raised != nil {
+		return nil, raised
+	}
+	var result *Object
+	if argc > 1 {
+		if args[1].typ == StrType {
+			return nil, f.RaiseType(TypeErrorType, "sum() can't sum strings [use ''.join(seq) instead]")
+		}
+		result = args[1]
+	} else {
+		result = NewInt(0).ToObject()
+	}
+	if result.typ == IntType {
+		item, raised := Next(f, iter)
+		for ; raised == nil; item, raised = Next(f, iter) {
+			if item.typ == IntType {
+				result = NewInt(toIntUnsafe(result).Value() + toIntUnsafe(item).Value()).ToObject()
+				continue
+			}
+			result, raised = Add(f, result, item)
+			break
+		}
+		if raised != nil {
+			if raised.isInstance(StopIterationType) {
+				f.RestoreExc(nil, nil)
+				return result, nil
+			}
+			return nil, raised
+		}
+	}
+	if result.typ == FloatType {
+		item, raised := Next(f, iter)
+		for ; raised == nil; item, raised = Next(f, iter) {
+			if item.typ == FloatType {
+				result = NewFloat(toFloatUnsafe(result).Value() + toFloatUnsafe(item).Value()).ToObject()
+				continue
+			}
+			if item.typ == IntType {
+				result = NewFloat(toFloatUnsafe(result).Value() + float64(toIntUnsafe(item).Value())).ToObject()
+				continue
+			}
+			result, raised = Add(f, result, item)
+			break
+		}
+		if raised != nil {
+			if raised.isInstance(StopIterationType) {
+				f.RestoreExc(nil, nil)
+				return result, nil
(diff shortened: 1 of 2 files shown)


runtime/builtin_types_test.go
```diff
@@ -433,3 +440,12 @@ func newTestIndexObject(index int) *Object {
...
 	return newObject(indexType)
 }
+
+func newTestAddObject(x int) *Object {
+	addType := newTestClass("Add", []*Type{ObjectType}, newStringDict(map[string]*Object{
+		"__add__": newBuiltinFunction("__add__", func(f *Frame, _ Args, _ KWArgs) (*Object, *BaseException) {
+			return NewInt(x + 1).ToObject(), nil
+		}).ToObject(),
+	}))
+	return newObject(addType)
```
trotterdylan: Rather than create a new helper, you can just create addType at the top of the test and then pass newObject(addType) to each of the relevant test cases. I try to avoid helpers unless they eliminate a significant amount of duplication.

runtime/builtin_types.go
```diff
@@ -617,6 +617,85 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		}
+		if raised != nil {
+			if raised.isInstance(StopIterationType) {
+				f.RestoreExc(nil, nil)
+				return result, nil
+			}
+			return nil, raised
+		}
+	}
+	if result.typ == FloatType {
```
trotterdylan: What's special about float and int (and not long?) that it can't all be the same logic?

runtime/builtin_types.go
```diff
@@ -617,6 +617,85 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		if args[1].typ == StrType {
+			return nil, f.RaiseType(TypeErrorType, "sum() can't sum strings [use ''.join(seq) instead]")
+		}
+		result = args[1]
+	} else {
+		result = NewInt(0).ToObject()
+	}
+	if result.typ == IntType {
+		item, raised := Next(f, iter)
+		for ; raised == nil; item, raised = Next(f, iter) {
```
trotterdylan: Is there one of the helpers in seq.go that can help here? seqForEach or seqApply come to mind.

aisk: I tried using seqForEach for these codes, but found i couldn't stop the iteration (while fount the next elem is not the same type like int or float). Looks like seqApply have the same limitation too?

runtime/builtin_types.go
aisk: This logic is from CPython, and i think it's trying to not create new python object in the middle while all the elements is int or float, and i think grumpy is having the same problem if we do not treat them special.

But i don't know should we treat longt type special like these, can you give me a suggestion?

runtime/builtin_types.go
trotterdylan: Ah, that makes sense. Thanks for the explanation.

runtime/builtin_types.go
```diff
@@ -650,6 +650,85 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			if item.typ == IntType {
+				result = NewInt(toIntUnsafe(result).Value() + toIntUnsafe(item).Value()).ToObject()
```
trotterdylan: `toIntUnsafe(result)` is not safe here because a prior call to Add() could have returned a non-int object.

runtime/builtin_types.go
trotterdylan: I see. I think we should prefer simplicity for now instead of optimizing at this state. Especially in light of the additional complexity that will be required to address my comment above about toIntUnsafe.

runtime/builtin_types.go
aisk: Hi, looks like L675 is assumed that result is an int type object, so i think it's ok here?

trotterdylan: But the call to `result, raised = Add(f, result, item)` may make result a non-int.

aisk: Sorry late for this commend, but i think there is a `break` under the `result, raised = Add(f, result, item` line, so this loop is breaked, and `toIntUnsafe` will not be called.

trotterdylan: Thanks for the explanation. I think I now properly understand the implementation: it iterates through while the sum is an int, then falls through to floats, and finally falls through to the general case. Previously I thought that each branch was mutually exclusive. It's a clever optimization.

runtime/builtin_types.go
trotterdylan: My confusion about toIntUnsafe I think demonstrates the down side of this optimization: the code is less understandable. In general I prefer to keep things as simple as possible and optimize only when there's a real performance problem. At that time, I'd want to see benchmarks that demonstrate the poor performance and concrete evidence that the optimization results in a significant improvement, and therefore justifies the additional complexity.

My recommendation is to make this function as simple as possible for now by simply doing Add() in a loop.

aisk:
Hi, i had another PR that implement this sum function just using `Add`: https://github.com/google/grumpy/pull/277

Maybe someday we want use this version of sum, so make another PR, and leave this closed.

PR #261, opened 2017-02-19
Implemented float.__hash__().

corona10:
For Implementing float.__hash__(), I referenced CPython and pypy implementations.
By implemented float hash, ordering of set and frozen set's elements between float type elements and other types elements are changed.

### DIFF
--- a/runtime/float.go
+++ b/runtime/float.go
@@ -20,6 +20,7 @@ import (
 	"math/big"
 	"reflect"
 	"strconv"
+	"sync"
 )
 
 // FloatType is the object representing the Python 'float' type.
@@ -28,12 +29,16 @@ var FloatType = newBasisType("float", reflect.TypeOf(Float{}), toFloatUnsafe, Ob
 // Float represents Python 'float' objects.
 type Float struct {
 	Object
-	value float64
+	value    float64
+	hashOnce sync.Once
+	hash     int
 }
 
 // NewFloat returns a new Float holding the given floating point value.
 func NewFloat(value float64) *Float {
-	return &Float{Object{typ: FloatType}, value}
+	result := Float{Object: Object{typ: FloatType}}
+	result.value = value
+	return &result
 }
 
 func toFloatUnsafe(o *Object) *Float {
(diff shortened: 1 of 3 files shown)


corona10:
@trotterdylan PTAL

runtime/float.go
```diff
@@ -28,12 +29,16 @@ var FloatType = newBasisType("float", reflect.TypeOf(Float{}), toFloatUnsafe, Ob
 // Float represents Python 'float' objects.
 type Float struct {
 	Object
-	value float64
+	value    float64
+	hashOnce sync.Once
```
trotterdylan: Once is heavyweight for this purpose since it relies on a mutex. I would recommend using an approach similar to what we do for str using atomic pointer operations.

runtime/float.go
```diff
@@ -28,12 +29,16 @@ var FloatType = newBasisType("float", reflect.TypeOf(Float{}), toFloatUnsafe, Ob
...
+	hash     int
 }
 
 // NewFloat returns a new Float holding the given floating point value.
 func NewFloat(value float64) *Float {
-	return &Float{Object{typ: FloatType}, value}
+	result := Float{Object: Object{typ: FloatType}}
```
trotterdylan: I assume this is changed because of the new struct member. What about using named field assignment like this:

```
return &Float{Object: Object{typ: FloatType}, value: value}
```

runtime/float.go
```diff
@@ -355,6 +374,39 @@ func floatDivModOp(f *Frame, method string, v, w *Object, fun func(v, w float64)
...
 
+func hashFloat(v float64) int {
+	if math.IsNaN(v) {
+		return 0
+	}
+
+	if math.IsInf(v, 0) {
+		if math.IsInf(v, 1) {
+			return 314159
+		} else if math.IsInf(v, -1) {
```
trotterdylan: When one branch returns, prefer to elide the else to minimize indentation: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

```
if math.IsInf(v, 1) {
  return 314159
}
if math.IsInf(v, -1) {
  return -271828
}
return 0
```

runtime/float.go
```diff
@@ -355,6 +374,39 @@ func floatDivModOp(f *Frame, method string, v, w *Object, fun func(v, w float64)
...
+			return -271828
+		} else {
+			return 0
+		}
+	} else {
```
trotterdylan: Similarly, elide the else since the if branch above always returns.

runtime/float.go
corona10: Thanks for reviews. I learn new techniques from you! It's nice. 👍

runtime/float.go
corona10: I should read this one! Thanks.

corona10:
@trotterdylan Thanks for your kind reviews, I updated it. PTAL.

Issue #17, opened 2017-01-04
Support tuple args in lambda


trotterdylan:
Per justinsaccount on HN:

$ make run
c = {}
top = sorted(c.items(), key=lambda (k,v): v)
^D
  Traceback (most recent call last):
    File "./tools/grumpc", line 102, in <module>
      sys.exit(main(parser.parse_args()))
    File "./tools/grumpc", line 60, in main
      visitor.visit(mod)
    File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 241, in visit
      return visitor(node)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/stmt.py", line 302, in visit_Module
      self._visit_each(node.body)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/stmt.py", line 632, in _visit_each
      self.visit(node)
    File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 241, in visit
      return visitor(node)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/stin visit_Assign
      with self.expr_visitor.visit(node.value) as value:
    File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 241, in visit
      return visitor(node)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/expr_visitor.py", line 101, in visit_Call
      values.append((util.go_str(k.arg), self.visit(k.value)))
    File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 241, in visit
      return visitor(node)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/expr_visitor.py", line 246, in visit_Lambda
      return self.visit_function_inline(func_node)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/expr_visitor.py", line 388, in visit_function_inline
      func_visitor = block.FunctionBlockVisitor(node)
    File "/Users/foo/src/grumpy/build/lib/python2.7/site-packages/grumpy/compiler/block.py", line 432, in __init__
[...]

PR #265, opened 2017-03-06
fix for #17

jamdagni86:
This is something which I could come up with in a couple of hours. The idea is to merge the constituent tuple arguments into a single argument (by joining argument names with τ) and get them unpacked inside the function using array notation.

Let me know your thoughts on the same. Unit tests are still work in progress; will update the PR once they are ready!

### DIFF
--- a/compiler/block.py
+++ b/compiler/block.py
@@ -249,6 +249,8 @@ class FunctionBlock(Block):
         if var:
           if var.type == Var.TYPE_GLOBAL:
             return self._resolve_global(writer, name)
+          if var.type == Var.TYPE_TUPLE_PARAM:
+            return expr.GeneratedLocalVar(name)
           writer.write_checked_call1('πg.CheckLocal(πF, {}, {})',
                                      util.adjust_local_name(name),
                                      util.go_str(name))
@@ -263,6 +265,7 @@ class Var(object):
   TYPE_LOCAL = 0
   TYPE_PARAM = 1
   TYPE_GLOBAL = 2
+  TYPE_TUPLE_PARAM = 3
 
   def __init__(self, name, var_type, arg_index=None):
     self.name = name
@@ -273,6 +276,9 @@ class Var(object):
     elif var_type == Var.TYPE_PARAM:
       assert arg_index is not None
       self.init_expr = 'πArgs[{}]'.format(arg_index)
+    elif var_type == Var.TYPE_TUPLE_PARAM:
+      assert arg_index is None
+      self.init_expr = 'nil'
     else:
       assert arg_index is None
       self.init_expr = None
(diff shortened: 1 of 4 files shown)


jamdagni86:
I signed it!

compiler/block.py
```diff
@@ -440,16 +440,29 @@ def __init__(self, node):
...
       args.append(node_args.vararg.arg)
     if node_args.kwarg:
       args.append(node_args.kwarg.arg)
     for i, name in enumerate(args):
-      if name in self.vars:
-        msg = "duplicate argument '{}' in function definition".format(name)
-        raise util.ParseError(node, msg)
-      self.vars[name] = Var(name, Var.TYPE_PARAM, arg_index=i)
+      if isinstance(name, list):
+        self.add_vars(i, 'τ'.join(name), node)
```
trotterdylan: As it turns out the arg tuples can be nested:

```
>>> def f((a, (b, c))):
...  print a, b, c
... 
>>> f((1, (2, 3)))
1 2 3
```

This points to a recursive approach similar to the way [_tie_target()](https://github.com/google/grumpy/blob/master/compiler/stmt.py#L751) in stmt.py works. My intuition is that we can use the exact same mechanism (grumpy.TieTarget struct) for unpacking the args.

Also, Var represents Python variables and are used to resolve names and handle scoping rules so I don't think you want to create a pseudo Var to represent the tuple. Instead you want to put the special unpacking logic somewhere around [here](https://github.com/google/grumpy/blob/master/compiler/expr_visitor.py#L464). But instead of assigning to each of the locals for the args, that's where you want to do the unpacking use the TieTarget mechanism.

jamdagni86: @trotterdylan - Since we are passing the Tuple as a single argument, we'll need some sort of Vars to hold the values inside the function as we need to **Tie** these arguments to the respective variables/arguments. Instead of pseudo Vars, what do you suggest?

Another approach could be to unpack them while constructing the arguments to the function itself and pass all the unpacked values as directly. 

Let me know!

trotterdylan: You can allocate a Go variable to hold the temporary value. Look for uses of Block.alloc_temp() in the code. The way this works for assignments is very similar to what I imagine you will need to do. It's a little indirect, but here's the relevant alloc for normal assignments: https://github.com/google/grumpy/blob/master/compiler/stmt.py#L640

I think the code for tuple argument unpacking will be very similar to that for tuple assignments. Hopefully you can reuse _build_assign_target(), _tie_target(), etc.

jamdagni86:
@trotterdylan - Haven't looked at the build failure yet; local make gofmt lint && make -j2 test didn't fail with this message.

Issue #269, opened 2017-03-19
Augmented assignment op not implemented: FloorDiv


luckcolors:
While trying to reselve dependencies again, i found this missing op.

From https://bitbucket.org/pypy/pypy/src/e8899d2b7da9ec497ebb892024237bb4f2d70867/lib-python/2.7/encodings/punycode.py?at=default&fileviewer=file-view-default
Line 95.

Thanks. :)

PR #270, opened 2017-03-21
Implement floordiv and NotImplemented.__repr__()

corona10:
1. Implement floordiv.
This PR is related with https://github.com/google/grumpy/issues/269
2. Implement NotImplemented.__repr__()


runtime/int.go
```diff
@@ -98,6 +98,22 @@ func intEq(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return intCompare(compareOpEq, toIntUnsafe(v), w), nil
 }
 
+func intFloorDiv(f *Frame, v, w *Object) (*Object, *BaseException) {
+	if !v.isInstance(IntType) {
+		return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("descriptor '__floordiv__' requires a 'int' object but received a '%s'", v.typ.Name()))
```
trotterdylan: The slot delegation code guarantees that the first argument to slot functions is of the correct type, so no need to check here.

runtime/int.go
```diff
@@ -98,6 +98,22 @@ func intEq(f *Frame, v, w *Object) (*Object, *BaseException) {
...
+	}
+
+	if w.isInstance(IntType) {
+		return intDivModOp(f, "__floordiv__", v, w, func(v, w int) (int, divModResult) {
+			return intCheckedDiv(v, w)
+		}, func(z, x, y *big.Int) {
+			longDiv(z, x, y)
+		})
+	}
+	fv, _ := intFloat(f, v)
```
trotterdylan: You should be able to return NotImplemented in this case and it will delegate to w's operator.

runtime/int.go
```diff
@@ -384,6 +400,7 @@ func initIntType(dict map[string]*Object) {
 	IntType.slots.Div = &binaryOpSlot{intDiv}
 	IntType.slots.DivMod = &binaryOpSlot{intDivMod}
 	IntType.slots.Eq = &binaryOpSlot{intEq}
+	IntType.slots.FloorDiv = &binaryOpSlot{intFloorDiv}
```
trotterdylan: Thinking about it a little more, I think you can just use intDiv for this slot. I think it does exactly what you want.

lib/operator.py
```diff
@@ -1,33 +1,405 @@
-def eq(a, b):
-  return a == b
+"""
```
trotterdylan: Seems like this change belongs in a different PR

corona10:
@trotterdylan Thanks for the early review. I will change them as your comment.

corona10:
@trotterdylan Ready for review. PTAL

runtime/int.go
```diff
@@ -471,6 +473,11 @@ func intCheckedDiv(v, w int) (int, divModResult) {
 	return q, r
 }
 
+func intCheckedFloorDiv(v, w int) (int, divModResult) {
```
trotterdylan: Is this still used somewhere?

corona10: @trotterdylan 
Not at all. Thanks, I didn't find that! I will update it

corona10:
@trotterdylan PTAL

PR #271, opened 2017-03-22
Added support for a+ and w+ modes (files)

yashpatel5400:
Supports missing modes of writing to files that are available in the Python standard libraries but not current supported here

### DIFF
--- a/runtime/file.go
+++ b/runtime/file.go
@@ -127,7 +127,12 @@ func fileInit(f *Frame, o *Object, args Args, _ KWArgs) (*Object, *BaseException
 		flag = os.O_RDONLY
 	case "r+", "r+b":
 		flag = os.O_RDWR
-	case "w", "wb":
+	// difference between r+ and a+ is that a+ automatically creates file
+	case "a+":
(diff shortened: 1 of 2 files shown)


runtime/file.go
```diff
@@ -127,6 +127,11 @@ func fileInit(f *Frame, o *Object, args Args, _ KWArgs) (*Object, *BaseException
 		flag = os.O_RDONLY
 	case "r+", "r+b":
 		flag = os.O_RDWR
+	// difference between r+ and a+ is that a+ automatically creates file
```
trotterdylan: Please use full sentences with capitalization and period for comments

trotterdylan:
This is great! Thanks for the PR. One minor coherent.

runtime/file.go
yashpatel5400: Thanks! Just pushed it with the correct formatting

corona10:
@yashpatel5400 FYI, `gofmt -w runtime/*.go` will help you for formatting problem 👍

PR #272, opened 2017-03-22
3rd party additions (from SVN source)

yashpatel5400:
Copied from http://svn.python.org/projects/python/branches/release27-maint/Lib/

### DIFF
--- a/runtime/file.go
+++ b/runtime/file.go
@@ -127,7 +127,12 @@ func fileInit(f *Frame, o *Object, args Args, _ KWArgs) (*Object, *BaseException
 		flag = os.O_RDONLY
 	case "r+", "r+b":
 		flag = os.O_RDWR
-	case "w", "wb":
+	// difference between r+ and a+ is that a+ automatically creates file
+	case "a+":
+		 flag = os.O_RDWR | os.O_CREATE | os.O_APPEND
+	case "w+":
+		 flag = os.O_RDWR | os.O_CREATE
+    case "w", "wb":
 		flag = os.O_WRONLY | os.O_CREATE | os.O_TRUNC
 	default:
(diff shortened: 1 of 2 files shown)


corona10:
@ yashpatel5400 Did you check all 3rd party works well?

corona10:
@yashpatel5400 And you looks like make a separate commit for this PR.

corona10:
I think you need to run tests like this way for your svn copied libraries.

https://github.com/google/grumpy/commit/d641069abd42a11325b2fee05c84ddb88b15a660

trotterdylan:
Please import CPython library sources into third_party/stdlib. Also, as Dong-hee says, everything that's checked into third_party should be confirmed to work. We've been doing this by copying libraries one-by-one from CPython or PyPy.

yashpatel5400:
So, should I create test files for each of the imported libraries and just move these into the stdlib? I didn't put them in stdlib initially because I'd be overwriting files, but I'll move them and do another request.

trotterdylan:
Many of the libraries already exist, so there's no need to import them. I recommend filing separate PRs for each library that you import. E.g. if you want to support mimetypes, then file a PR with mimetypes.py and test/test_mimetypes.py.

corona10:
@yashpatel5400 
To use 3rd party libraries on grumpy needs to implement basic python operation some of them are not implemented yet and some of them are not 100% compatible.

So what you need to do is importIng each of 3rd party library and then run unit tests for them.
Each of unit tests should be worked well on grumpy. If not you should implement related API or modify that API.

yashpatel5400:
Actually,

yashpatel5400:
@corona10 Ah, got it. I'll retract the current push, since it's just all the source files without any tests written for them and start to do separate ones for each

PR #277, opened 2017-03-26
implement a simple builtin sum

aisk:


runtime/builtin_types.go
```diff
@@ -662,6 +662,32 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		if args[1].typ == StrType {
+			return nil, f.RaiseType(TypeErrorType, "sum() can't sum strings [use ''.join(seq) instead]")
+		}
+		result = args[1]
+	} else {
+		result = NewInt(0).ToObject()
+	}
+	raised := seqForEach(f, args[0], func(o *Object) (raised *BaseException) {
+		result, raised = Add(f, result, o)
+		return
```
trotterdylan: Please explicitly list what's being returned: `return raised`

runtime/builtin_types.go
```diff
@@ -662,6 +662,32 @@ func builtinSorted(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		result = args[1]
+	} else {
+		result = NewInt(0).ToObject()
+	}
+	raised := seqForEach(f, args[0], func(o *Object) (raised *BaseException) {
+		result, raised = Add(f, result, o)
+		return raised
+	})
+
+	return result, raised
```
trotterdylan: Sorry, one more suggestion. Check for raised and return nil, raised if it's set. In general, only one of the two return values should be set so that callers don't get confused.

```
if raised != nil {
  return nil, raised
}
return result, nil

aisk:
@trotterdylan Hi this can be reviewed now

PR #278, opened 2017-03-29
Import stdlib colorsys and implement round function for ndigits.

corona10:
Let's support `colorsys`.
Now grumpy passed unittest `test_colorsys.py` fully.


runtime/builtin_types_test.go
```diff
@@ -293,7 +293,10 @@ func TestBuiltinFuncs(t *testing.T) {
 		{f: "round", args: wrapArgs(-1234.111), want: NewFloat(-1234).ToObject()},
 		{f: "round", args: wrapArgs(1234.567, newTestIndexObject(0)), want: NewFloat(1235).ToObject()},
 		{f: "round", args: wrapArgs("foo"), wantExc: mustCreateException(TypeErrorType, "a float is required")},
-		{f: "round", args: wrapArgs(1234.5, 1), wantExc: mustCreateException(NotImplementedErrorType, "round with ndigits is not implemented in grumpy")},
+		{f: "round", args: wrapArgs(1234.5, 1), want: NewFloat(1234.5).ToObject()},
+		{f: "round", args: wrapArgs(1234.5, 1), want: NewFloat(1234.5).ToObject()},
+		{f: "round", args: wrapArgs(1234.56, 1), want: NewFloat(1234.6).ToObject()},
+		{f: "round", args: wrapArgs(-1234.56, 1), want: NewFloat(-1234.6).ToObject()},
```
aisk: Looks like we should implement the round function with second parameter as a negative number.

corona10: @aisk Yes, you are right. `round(-1234.5, -1)` should be a `-1230.`

runtime/builtin_types.go
```diff
@@ -631,15 +631,23 @@ func builtinRound(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		return NewFloat(number).ToObject(), nil
+	}
+
+	var pow float64 = 1
+	var result float64
+	for i := 0; i < ndigits; i++ {
+		pow *= 10
+	}
+
+	result = float64(math.Floor(number*pow+0.5)) / pow
```
trotterdylan: Is the float64(...) necessary here? The return value of Floor should already by float64.

runtime/builtin_types.go
```diff
@@ -631,15 +631,23 @@ func builtinRound(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
 		return nil, f.RaiseType(TypeErrorType, "a float is required")
 	}
-	return NewFloat(math.Floor(number + 0.5)).ToObject(), nil
+
+	if math.IsNaN(number) || math.IsInf(number, 0) || number == 0.0 {
+		return NewFloat(number).ToObject(), nil
+	}
+
+	var pow float64 = 1
+	var result float64
```
trotterdylan: No nee to declare this, just use `result := ...` down below.

runtime/builtin_types.go
```diff
@@ -631,15 +631,23 @@ func builtinRound(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
 	if !isFloat {
 		return nil, f.RaiseType(TypeErrorType, "a float is required")
 	}
-	return NewFloat(math.Floor(number + 0.5)).ToObject(), nil
+
+	if math.IsNaN(number) || math.IsInf(number, 0) || number == 0.0 {
+		return NewFloat(number).ToObject(), nil
+	}
+
+	var pow float64 = 1
```
trotterdylan: `pow := 1.0`

Go should be able to infer it's a float64 based on its usage below.

runtime/builtin_types_test.go
corona10: @aisk 
Now I add a test with the second parameter as a negative number.
And it works!

corona10:
@trotterdylan PTAL

runtime/builtin_types.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	//	"strconv"
```
trotterdylan: Remove

runtime/builtin_types.go
```diff
@@ -864,6 +886,15 @@ func initIters(f *Frame, items []*Object) ([]*Object, *BaseException) {
 	return iters, nil
 }
 
+func round(x float64, ndigits int) float64 {
+	pow := 1.0
+	for i := 0; i < ndigits; i++ {
+		pow *= 10
+	}
+	rounded := math.Floor(x*pow+0.5) / pow
```
trotterdylan: return math.Floor(...

runtime/builtin_types_test.go
trotterdylan: Probably worth adding another couple negative test cases.

runtime/builtin_types.go
```diff
@@ -631,15 +632,36 @@ func builtinRound(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
-	return NewFloat(math.Floor(number + 0.5)).ToObject(), nil
+
+	if math.IsNaN(number) || math.IsInf(number, 0) || number == 0.0 {
+		return NewFloat(number).ToObject(), nil
+	}
+
+	var result float64
+	if ndigits >= 0 {
+		result = round(number, ndigits)
+		if math.Abs(result-number) == 0.5 {
```
trotterdylan: I don't understand this part. Can you add a comment describing what you're doing here?

corona10: @trotterdylan 
You mean
```
                if math.Abs(result-number) == 0.5 {
                        if number > 0.0 {
                                result = number + 0.5
                        } else {
                                result = number - 0.5
                        }
                }
```
this part?

trotterdylan: Yeah

runtime/builtin_types.go
```diff
@@ -631,15 +631,39 @@ func builtinRound(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
-	return NewFloat(math.Floor(number + 0.5)).ToObject(), nil
+
+	if math.IsNaN(number) || math.IsInf(number, 0) || number == 0.0 {
+		return NewFloat(number).ToObject(), nil
+	}
+
+	var result float64
+	if ndigits == 0 {
+		result = round(number, ndigits)
+		// round(12.5) => 13, round(-12.5) => -13
```
corona10: @trotterdylan 
That part intended to do this.

corona10:
Updated!

runtime/builtin_types.go
trotterdylan: My intuition is that negative ndigits should "just work" with the normal logic because we're using pow.

Additionally, my intuition about negative values for `number` is that we should just be able to do the same logic we do for positive values except on `-number`.

What I'm hoping for from a comment is to help readers understand why it's so much more complicated than that.

corona10:
````
/**
In Python 2, ndigits <0 treats the decimal point as .0, and the integer as | Ndigits | It is masked to 0 as | Digits |.
To do this, divide -12.5 by pow (10, | ndigits |) (this will be -1.25) and round it. Finally, multiplying by pow (10, | ndigits |) yields -10.0.
**/

````
@trotterdylan 
Will be a good enough?

Comment (trotterdylan) [inline runtime/builtin_types.go]: I guess I still don't understand why we need all the special cases. ISTM that something like this will work for all cases:

```
neg := false
if number < 0 {
	neg = true
	number = -number
}
pow := math.Pow(10.0, float64(ndigits))
result = math.Floor(number*pow+0.5) / pow
if neg {
	result = -result
}
```

corona10:
@trotterdylan You are awesome! 👍 It's simpler!

PR #280, opened 2017-04-05
[WIP] Implement complex.__new__

m4ns0ur:


runtime/complex.go
```diff
@@ -68,5 +70,190 @@ func complexRepr(f *Frame, o *Object) (*Object, *BaseException) {
...
+// <float><signed-float>j, <float><sign>j, <sign>j or j, where <float> is
+// any numeric string that's acceptable by strconv.ParseFloat(s, 64)),
+// ParseComplex returns the respective complex128 number.
+func parseComplex(s string) (complex128, error) {
+	ts := strings.Trim(s, "() ")
+	if !strings.Contains(ts, "j") {
+		result, err := strconv.ParseFloat(ts, 64)
+		return complex(result, 0), err
+	}
+	ts = strings.Replace(ts, "j", "", -1)
```
trotterdylan: It's possible this would be cleaner by parsing the expression as a regular expression. According to the [CPython source](https://github.com/python/cpython/blob/2.7/Objects/complexobject.c#L1021), the pattern is basically `<float>|<float>j|<float><signed float>j|<float>[+-]j|[+-]j|j` where floats have the form given here: https://docs.python.org/2/reference/lexical_analysis.html#floating-point-literals

It's not exactly simple, but I think the code would be easier to reason about. As it stands, it's pretty difficult to understand what's going on here.

runtime/complex_test.go
```diff
@@ -31,3 +215,83 @@ func TestComplexRepr(t *testing.T) {
...
+		{"-1.2-3.4", []string{"", "-1.2", "-3.4"}},
+	}
+	for _, cas := range cases {
+		if got := splitBeforeSign(cas.s); !eq(got, cas.want) {
+			t.Errorf("splitBeforeSign(%q) = %v, want %v", cas.s, got, cas.want)
+		}
+	}
+}
+
+func eq(a, b []string) bool {
```
trotterdylan: Please choose a different name for this. Perhaps stringSlicesAreEq()

trotterdylan:
Note that the Travis failure is due to a flakey test and not your code. I've spent some time trying to track it down, but it's kind of gnarly. There's a minor race condition in the raw_input test relating to writing to stdin. I think it's a test-only issue.

m4ns0ur:
making separate PR for complex comparison

runtime/complex.go
m4ns0ur: I'm fixing this in separate commit

m4ns0ur:
@trotterdylan sorry! I shouldn't rebase with master, now there is an unrelated commit in this thread history, and googlebot is complaining 😕
shall I create a clean PR, or you will squash it?

trotterdylan:
I think it'll be easier to create a new PR so that googlebot doesn't get confused. Thanks!

PR #281, opened 2017-04-06
Implement complex comparison operations

m4ns0ur:


corona10:
@trotterdylan 
TestRawInput` looks like have a significant issue. 
I will also take a look into this issue.

runtime/complex.go
```diff
@@ -68,5 +69,80 @@ func complexRepr(f *Frame, o *Object) (*Object, *BaseException) {
...
+
+func complexEq(f *Frame, v, w *Object) (*Object, *BaseException) {
+	e, ok := complexCompare(toComplexUnsafe(v), w)
+	if !ok {
+		return NotImplemented, nil
+	}
+	return GetBool(e).ToObject(), nil
+}
+
+func complexGE(f *Frame, v, w *Object) (*Object, *BaseException) {
```
trotterdylan: Rather than writing a wrapper function for each of complexLE, GE, etc. I would just make comparisonNotSupported compatible with binary op and then use it for all the unsupported slots:

```
ComplexType.slots.GE = &binaryOpSlot{comparisonNotSupported}
ComplexType.slots.GT = &binaryOpSlot{comparisonNotSupported}
...
```

runtime/complex.go
```diff
@@ -68,5 +69,80 @@ func complexRepr(f *Frame, o *Object) (*Object, *BaseException) {
...
+func complexCompare(v *Complex, w *Object) (bool, bool) {
+	lhsr := real(v.Value())
+	rhs, ok := complexCoerce(w)
+	if !ok {
+		return false, false
+	}
+	return lhsr == real(rhs) && imag(v.Value()) == imag(rhs), true
+}
+
+func comparisonNotSuported(f *Frame, w *Object) (*Object, *BaseException) {
```
trotterdylan: Maybe call this complexCompareNotSupported or something, since comparisonNotSupported is pretty generic.

runtime/complex.go
m4ns0ur: good point

m4ns0ur:

PR #293, opened 2017-04-23
Implement complex hash.

corona10:
Implement complex hash.

### DIFF
--- a/runtime/complex.go
+++ b/runtime/complex.go
@@ -72,6 +72,7 @@ func initComplexType(dict map[string]*Object) {
 	ComplexType.slots.Eq = &binaryOpSlot{complexEq}
 	ComplexType.slots.GE = &binaryOpSlot{complexCompareNotSupported}
 	ComplexType.slots.GT = &binaryOpSlot{complexCompareNotSupported}
+	ComplexType.slots.Hash = &unaryOpSlot{complexHash}
 	ComplexType.slots.LE = &binaryOpSlot{complexCompareNotSupported}
 	ComplexType.slots.LT = &binaryOpSlot{complexCompareNotSupported}
 	ComplexType.slots.NE = &binaryOpSlot{complexNE}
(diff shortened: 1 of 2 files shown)


corona10:
@trotterdylan Ready for review.

runtime/complex.go
```diff
@@ -28,11 +30,12 @@ var ComplexType = newBasisType("complex", reflect.TypeOf(Complex{}), toComplexUn
 type Complex struct {
 	Object
 	value complex128
+	hash  int
```
trotterdylan: I question whether caching the hash value has much value. Note that CPython does not cache it for complex or float. Perhaps we should remove the caching in float until there's a good reason to add it.

Note that strings are good candidates for caching since a) the length of the string can be very long (amortizes the runtime of hash(s)) and b) many commonly used strings (e.g. attribute names) are interned meaning that the cache is highly effective.

Neither of these factors apply to float or complex.

corona10: @trotterdylan Yes, I agree with you. It is not a heavy operation to caching it.

corona10:
@trotterdylan PTAL

runtime/complex.go
```diff
@@ -32,7 +32,7 @@ type Complex struct {
 
 // NewComplex returns a new Complex holding the given complex value.
 func NewComplex(value complex128) *Complex {
-	return &Complex{Object{typ: ComplexType}, value}
+	return &Complex{Object: Object{typ: ComplexType}, value: value}
```
trotterdylan: This change is no longer necessary

runtime/complex.go
```diff
@@ -130,3 +131,14 @@ func complexCoerce(o *Object) (complex128, bool) {
 	}
 	return complex(floatO, 0.0), true
 }
+
+func complexHash(f *Frame, o *Object) (*Object, *BaseException) {
+	v := toComplexUnsafe(o)
+	hashreal := hashFloat(real(v.Value()))
```
trotterdylan: Style nit: I don't think you can reduce the number of temporaries without affecting readability too much.

corona10:
@trotterdylan Thank you for your review! I updated with you pointed it.

PR #294, opened 2017-04-23
Implement Complex.Add and Sub

corona10:
* Now grumpy supports such as 1+3j or 2-7j.
* Add unittests for binary operations.
* complex tests of test_operator.py is now passed.

### DIFF
--- a/runtime/complex.go
+++ b/runtime/complex.go
@@ -49,6 +49,43 @@ func (c *Complex) Value() complex128 {
 	return c.value
 }
 
+func complexAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
+	return complexArithmeticOp(f, "__add__", v, w, func(lhs, rhs complex128) complex128 {
+		return lhs + rhs
+	})
+}
+
+func complexEq(f *Frame, v, w *Object) (*Object, *BaseException) {
+	e, ok := complexCompare(toComplexUnsafe(v), w)
+	if !ok {
+		return NotImplemented, nil
+	}
+	return GetBool(e).ToObject(), nil
+}
+
+func complexHash(f *Frame, o *Object) (*Object, *BaseException) {
+	v := toComplexUnsafe(o).Value()
+	hashCombined := hashFloat(real(v)) + 1000003*hashFloat(imag(v))
+	if hashCombined == -1 {
+		hashCombined = -2
+	}
+	return NewInt(hashCombined).ToObject(), nil
+}
+
+func complexNE(f *Frame, v, w *Object) (*Object, *BaseException) {
+	e, ok := complexCompare(toComplexUnsafe(v), w)
+	if !ok {
+		return NotImplemented, nil
+	}
+	return GetBool(!e).ToObject(), nil
+}
+
+func complexRAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
+	return complexArithmeticOp(f, "__radd__", v, w, func(lhs, rhs complex128) complex128 {
+		return lhs + rhs
+	})
+}
+
 func complexRepr(f *Frame, o *Object) (*Object, *BaseException) {
 	c := toComplexUnsafe(o).Value()
 	rs, is := "", ""
@@ -68,7 +105,20 @@ func complexRepr(f *Frame, o *Object) (*Object, *BaseException) {
 	return NewStr(fmt.Sprintf("%s%s%s%sj%s", pre, rs, sign, is, post)).ToObject(), nil
 }
 
+func complexRSub(f *Frame, v, w *Object) (*Object, *BaseException) {
+	return complexArithmeticOp(f, "__rsub__", v, w, func(lhs, rhs complex128) complex128 {
+		return rhs - lhs
+	})
+}
+
+func complexSub(f *Frame, v, w *Object) (*Object, *BaseException) {
+	return complexArithmeticOp(f, "__rsub__", v, w, func(lhs, rhs complex128) complex128 {
+		return lhs - rhs
+	})
+}
+
 func initComplexType(dict map[string]*Object) {
+	ComplexType.slots.Add = &binaryOpSlot{complexAdd}
 	ComplexType.slots.Eq = &binaryOpSlot{complexEq}
 	ComplexType.slots.GE = &binaryOpSlot{complexCompareNotSupported}
 	ComplexType.slots.GT = &binaryOpSlot{complexCompareNotSupported}
@@ -76,23 +126,10 @@ func initComplexType(dict map[string]*Object) {
 	ComplexType.slots.LE = &binaryOpSlot{complexCompareNotSupported}
 	ComplexType.slots.LT = &binaryOpSlot{complexCompareNotSupported}
 	ComplexType.slots.NE = &binaryOpSlot{complexNE}
+	ComplexType.slots.RAdd = &binaryOpSlot{complexRAdd}
 	ComplexType.slots.Repr = &unaryOpSlot{complexRepr}
-}
-
-func complexEq(f *Frame, v, w *Object) (*Object, *BaseException) {
-	e, ok := complexCompare(toComplexUnsafe(v), w)
-	if !ok {
-		return NotImplemented, nil
-	}
-	return GetBool(e).ToObject(), nil
-}
-
-func complexNE(f *Frame, v, w *Object) (*Object, *BaseException) {
-	e, ok := complexCompare(toComplexUnsafe(v), w)
-	if !ok {
-		return NotImplemented, nil
-	}
-	return GetBool(!e).ToObject(), nil
+	ComplexType.slots.RSub = &binaryOpSlot{complexRSub}
+	ComplexType.slots.Sub = &binaryOpSlot{complexSub}
 }
 
 func complexCompare(v *Complex, w *Object) (bool, bool) {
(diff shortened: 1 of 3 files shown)


corona10:
@trotterdylan 
I have no idea this implementation is a right way to do.
Because it looks like in CPython handling `3+4j` in this way.
````
  2           0 LOAD_CONST               3 ((3+4j))
              3 STORE_FAST               0 (a)

````
But in the pythonparser handling the complex as below
````
LOAD_CONST 3
STORE_FAST 0
LOAD_CONST 4j
STORE_FAST 1

LOAD_FAST 0
LOAD_FAST 1
BINARY_ADD
STORE_FAST 0
````
Please let I know if I think in a wrong way.

File: runtime/complex.go [hunk 1]
```diff
@@ -130,3 +131,40 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+	if w.isInstance(FloatType) {
+		return complexArithmeticOp(f, "__add__", v, w, complexAddFloat)
+	}
+	return NotImplemented, nil
+}
+
+func complexAddComplex(v, w *Object) complex128 {
+	lhs := toComplexUnsafe(v).Value()
+	rhs := toComplexUnsafe(w).Value()
+	return complex(real(lhs)+real(rhs), imag(lhs)+imag(rhs))
```
trotterdylan: Go supports complex addition, you can just do: `return lhs + rhs`

Because this is so simple and it's only used in one place you can just do this operation in line and remove the complexAddComplex function.

runtime/complex.go
```diff
@@ -130,3 +131,40 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+func complexArithmeticOp(f *Frame, method string, v, w *Object, fun func(v, w *Object) complex128) (*Object, *BaseException) {
+	return NewComplex(fun(v, w)).ToObject(), nil
+}
+
+func complexAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
+	if w.isInstance(ComplexType) {
+		return complexArithmeticOp(f, "__add__", v, w, complexAddComplex)
+	}
+
+	if w.isInstance(IntType) {
```
trotterdylan: I think if w is not complex then it's cleaner to use floatCoerce(w). That will also support long properly.

runtime/float.go
```diff
@@ -59,6 +59,9 @@ func floatAbs(f *Frame, o *Object) (*Object, *BaseException) {
 }
 
 func floatAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
+	if w.isInstance(ComplexType) {
```
trotterdylan: I believe the way that CPython works is to let float.\_\_add\_\_ return NotImplemented and then fall back to complex.\_\_radd\_\_. That's how coercion works for other numeric types. So if you implement \_\_radd\_\_ for complex, you can remove these bits of floatAdd and intAdd.

corona10:
@trotterdylan Ready for review man!

runtime/complex.go
```diff
@@ -140,3 +144,91 @@ func complexHash(f *Frame, o *Object) (*Object, *BaseException) {
 	}
 	return NewInt(hashCombined).ToObject(), nil
 }
+
+func complexArithmeticOpComplex(f *Frame, method string, v, w complex128, fun func(v, w complex128) complex128) (*Object, *BaseException) {
+	return NewComplex(fun(v, w)).ToObject(), nil
+}
+
+func complexArithmeticOp(f *Frame, method string, v complex128, w float64, fun func(v complex128, w float64) complex128) (*Object, *BaseException) {
```
trotterdylan: Seems like you could fold all the coercion and overflow checking logic into this function so you don't have to duplicate it for every operator.

runtime/complex.go
```diff
@@ -140,3 +144,91 @@ func complexHash(f *Frame, o *Object) (*Object, *BaseException) {
 	}
 	return NewInt(hashCombined).ToObject(), nil
 }
+
+func complexArithmeticOpComplex(f *Frame, method string, v, w complex128, fun func(v, w complex128) complex128) (*Object, *BaseException) {
```
trotterdylan: Please move helper functions to the end of the file and keep method functions (e.g. complexAdd, complexSub) in alphabetical order below the initComplexType() function.

runtime/complex_test.go
```diff
@@ -42,6 +43,36 @@ func TestComplexEq(t *testing.T) {
...
+		{Add, NewInt(3).ToObject(), NewComplex(3i).ToObject(), NewComplex(3 + 3i).ToObject(), nil},
+		{Add, NewLong(big.NewInt(9999999)).ToObject(), NewComplex(3i).ToObject(), NewComplex(9999999 + 3i).ToObject(), nil},
+		{Add, NewFloat(3.5).ToObject(), NewComplex(3i).ToObject(), NewComplex(3.5 + 3i).ToObject(), nil},
+		{Sub, NewComplex(1 + 3i).ToObject(), NewComplex(1 + 3i).ToObject(), NewComplex(0i).ToObject(), nil},
+		{Sub, NewComplex(1 + 3i).ToObject(), NewComplex(3i).ToObject(), NewComplex(1).ToObject(), nil},
+		{Sub, NewComplex(1 + 3i).ToObject(), NewFloat(1).ToObject(), NewComplex(3i).ToObject(), nil},
+		{Sub, NewComplex(3i).ToObject(), NewFloat(1.2).ToObject(), NewComplex(-1.2 + 3i).ToObject(), nil},
+		{Sub, NewComplex(1 + 3i).ToObject(), NewComplex(1 + 3i).ToObject(), NewComplex(0i).ToObject(), nil},
+		{Sub, NewComplex(4 + 3i).ToObject(), NewInt(1).ToObject(), NewComplex(3 + 3i).ToObject(), nil},
+		{Sub, NewComplex(4 + 3i).ToObject(), NewLong(big.NewInt(99994)).ToObject(), NewComplex(-99990 + 3i).ToObject(), nil},
```
trotterdylan: We should have some inf and nan tests. Also some test cases that raise exceptions.

corona10:
@trotterdylan 
Nice review!
I refactored some functions. Please take a look.
Writing a test function for a NaN case is a bit tricky. What do you think?
```
>>> c == (float('nan') + 3j)
False
>>> c
(nan+3j)
```

runtime/complex.go
```diff
@@ -49,6 +49,7 @@ func (c *Complex) Value() complex128 {
 	return c.value
 }
 
+// TODO: Move this function below the initComplexType()
```
trotterdylan: Sorry! I gave you the wrong advice, slot and method functions should be above the initXXX() function and helpers (e.g. complexArithmeticOp) should be below.

runtime/complex.go
```diff
@@ -140,3 +161,42 @@ func complexHash(f *Frame, o *Object) (*Object, *BaseException) {
...
+	}
+
+	floatW, ok := floatCoerce(w)
+	if !ok {
+		if math.IsInf(floatW, 0) {
+			return nil, f.RaiseType(OverflowErrorType, "long int too large to convert to float")
+		}
+		return NotImplemented, nil
+	}
+	return NewComplex(fun2(toComplexUnsafe(v).Value(), floatW)).ToObject(), nil
```
trotterdylan: I would just convert floatW in this case to a complex(floatW, 0) and call fun1 in all cases. Then you can get rid of the fun2 param altogether.

runtime/complex.go
```diff
@@ -95,6 +104,18 @@ func complexNE(f *Frame, v, w *Object) (*Object, *BaseException) {
 	return GetBool(!e).ToObject(), nil
 }
 
+func complexRAdd(f *Frame, v, w *Object) (*Object, *BaseException) {
+	return complexArithmeticOp(f, "__radd__", v, w, complexAddComplex, complexAddFloat)
+}
+
+func complexRSub(f *Frame, v, w *Object) (*Object, *BaseException) {
+	return complexArithmeticOp(f, "__rsub__", v, w, complexRSubComplex, complexRSubFloat)
```
trotterdylan: Since the complex*SubComplex() functions are only called in one place, I would just define the function inline here:

```
return complexArithmeticOp(f, "__rsub__", v, w, func(v, w complex128) complex128 {
  ...
})
```

corona10:
@trotterdylan Awesome! I removed unnecessary parts and move functions positions.

trotterdylan:
NaN needs to be handled specially. See TestFloatArithmeticOps: https://github.com/google/grumpy/blob/master/runtime/float_test.go#L105

corona10:
@trotterdylan Done.

corona10:
PTAL

runtime/complex_test.go
```diff
@@ -42,6 +44,68 @@ func TestComplexEq(t *testing.T) {
...
+		{Sub, NewFloat(math.Inf(1)).ToObject(), NewComplex(3i).ToObject(), NewComplex(complex(math.Inf(1), -3)).ToObject(), nil},
+		{Sub, NewFloat(math.Inf(-1)).ToObject(), NewComplex(3i).ToObject(), NewComplex(complex(math.Inf(-1), -3)).ToObject(), nil},
+		{Sub, NewComplex(1 + 3i).ToObject(), None, nil, mustCreateException(TypeErrorType, "unsupported operand type(s) for -: 'complex' and 'NoneType'")},
+		{Sub, None, NewComplex(1 + 3i).ToObject(), nil, mustCreateException(TypeErrorType, "unsupported operand type(s) for -: 'NoneType' and 'complex'")},
+		{Sub, NewFloat(math.NaN()).ToObject(), NewComplex(3i).ToObject(), NewComplex(complex(math.NaN(), -3)).ToObject(), nil},
+		{Sub, NewComplex(cmplx.NaN()).ToObject(), NewComplex(3i).ToObject(), NewComplex(cmplx.NaN()).ToObject(), nil},
+		{Sub, NewFloat(math.Inf(-1)).ToObject(), NewComplex(complex(math.Inf(-1), 3)).ToObject(), NewComplex(complex(math.NaN(), -3)).ToObject(), nil},
+	}
+
+	NaNCheck := func(a, b float64) bool {
```
trotterdylan: I'm not sure about this, but it might be cleaner to have a helper for floats:

```
func floatsAreSame(a, b float64) bool {
  return a == b || (math.IsNaN(a) && math.IsNaN(b))
}

func complexesAreSame(a, b complex128) bool {
  return floatsAreSame(real(a), real(b)) && floatsAreSame(imag(a), imag(b))
}
```

I think these will be useful in the other places we do this kind of comparison (float_test.go and in https://github.com/google/grumpy/pull/296). The checks will then be much simpler:

```
if got == nil || cas.want == nil || !got.isInstance(ComplexType) || !cas.want.isInstance(ComplexType) || !complexesAreSame(toComplexUnsafe(got).Value(), toComplexUnsafe(cas.want).Value()) {
        t.Errorf("%s(%v, %v) = %v, want %v", getFuncName(cas.fun), cas.v, cas.w, got, cas.want)
}
```

corona10: @trotterdylan You are great!

Issue #262, opened 2017-02-19
float.__str__ should return only 12 digits of precision


corona10:
`math.ldexp(0.125,1000)` should be `1.33938575898e+300`.
But grumpy results `1.3393857589828342e+300`.

trotterdylan:
For CPython I get:

```
$ python
>>> math.ldexp(0.125,1000)
1.3393857589828342e+300
```

And for Grumpy:

```
$ echo 'import math; print math.ldexp(0.125,1000)' | make run
1.3393857589828342e+300
```

Where I do see your result is if I run:

```
python -c 'import math; print math.ldexp(0.125,1000)'
1.33938575898e+300
```

This has to do with the difference between `__str__` and `__repr__` for float. `__str__` seems to truncate a bit, whereas `__repr__` shows the exact result. From the CPython source:

```
static PyObject *
float_repr(PyFloatObject *v)
{
    return float_str_or_repr(v, 0, 'r');
}

static PyObject *
float_str(PyFloatObject *v)
{
    return float_str_or_repr(v, PyFloat_STR_PRECISION, 'g');
}
```

This in turn results in different behavior in PyOS_double_to_string() where repr uses 17 digits of precision (guaranteed to be able to represent the number exactly) and str uses `PyFloat_STR_PRECISION == 12` digits of precision.

So really the bug here is that Grumpy has no `float.__str__` and so falls back to `float.__repr__`. We should implement `float.__str__` to return only 12 digits of precision.

corona10:
@trotterdylan Yes, you are right. 👍

corona10:
@trotterdylan `float_str_or_repr` should be reimplementeed in python way.
When I read Cpython's implementations, it is little bits different between golang's `strconv.FormatFloat`.

PR #296, opened 2017-04-24
Implement complex.__new__ and complex.__neg__

m4ns0ur:


runtime/complex.go
```diff
@@ -130,3 +134,130 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format, result.typ.Name()))
+			}
+			return result, nil
+		}
+		if floatSlot := o.typ.slots.Float; floatSlot != nil {
+			result, raised := floatConvert(floatSlot, f, o)
+			if raised != nil {
+				return nil, raised
+			}
+			f := toFloatUnsafe(result).Value()
```
trotterdylan: If floatConvert() returned (*Float, *BaseException) then this unsafe cast would be unnecessary.

runtime/complex.go
```diff
@@ -130,3 +134,130 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+// If string is well-formed (one of these forms: <float>, <float>j,
+// <float><signed-float>j, <float><sign>j, <sign>j or j, where <float> is
+// any numeric string that's acceptable by strconv.ParseFloat(s, 64)),
+// ParseComplex returns the respective complex128 number.
+func parseComplex(s string) (complex128, error) {
+	c := strings.Count(s, "(")
+	if (c > 1) || (c == 1 && strings.Count(s, ")") != 1) {
+		return complex(0, 0), errors.New("Malformed complex string, more than one matching parantheses")
+	}
+	ts := strings.Trim(s, "() ")
```
trotterdylan: Note that all whitespace is allowed around the parentheses:

```
>>> complex(' (\n1+2j\t)  \t ')
(1+2j)
```

So maybe the thing to do is to TrimSpace(s), remove parens from the beginning and end and then do another TrimSpace() on the inside of the parens.

runtime/complex.go
```diff
@@ -130,3 +134,130 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+	sj := `(?P<sign5>[-+])j`
+	j := `(?P<onlyJ>j)`
+	r := regexp.MustCompile(`^(?:` + fsfj + `|` + fsj + `|` + fj + `|` + f + `|` + sj + `|` + j + `)$`)
+	subs := r.FindStringSubmatch(ts)
+	if subs == nil {
+		return complex(0, 0), errors.New("Malformed complex string, no mathing pattern found")
+	}
+	res := make(map[string]string)
+	for i, name := range r.SubexpNames() {
+		// First one is the complete string
```
trotterdylan: Comments should be full sentences (including period).

runtime/complex.go
```diff
@@ -130,3 +134,130 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+	fj := `(?P<imag3>` + fre + `)j`
+	f := `(?P<real4>` + fre + `)`
+	sj := `(?P<sign5>[-+])j`
+	j := `(?P<onlyJ>j)`
+	r := regexp.MustCompile(`^(?:` + fsfj + `|` + fsj + `|` + fj + `|` + f + `|` + sj + `|` + j + `)$`)
+	subs := r.FindStringSubmatch(ts)
+	if subs == nil {
+		return complex(0, 0), errors.New("Malformed complex string, no mathing pattern found")
+	}
+	res := make(map[string]string)
```
trotterdylan: Allocation in Go is relatively expensive. I think you can avoid allocating this map by using indices for the sub-matches and using a fixed length array: var res [6]string. Constants for the index names would make it more readable.

runtime/complex_test.go
```diff
@@ -80,6 +83,92 @@ func TestComplexNE(t *testing.T) {
...
+	}
+	subType := newTestClass("SubType", []*Type{ComplexType}, newStringDict(map[string]*Object{}))
+	subTypeObject := (&Complex{Object: Object{typ: subType}, value: 3.14}).ToObject()
+	slotSubTypeType := newTestClass("SlotSubType", []*Type{ObjectType}, newStringDict(map[string]*Object{
+		"__complex__": newBuiltinFunction("__complex__", func(_ *Frame, _ Args, _ KWArgs) (*Object, *BaseException) {
+			return subTypeObject, nil
+		}).ToObject(),
+	}))
+	cases := []invokeTestCase{
+		{args: wrapArgs(ComplexType), want: NewComplex(0).ToObject()},
```
trotterdylan: Love all these test cases! Thanks for being so thorough.

runtime/complex.go
```diff
@@ -130,3 +134,133 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+				return nil, f.RaiseType(TypeErrorType, fmt.Sprintf(format, result.typ.Name()))
+			}
+			return result, nil
+		}
+		if floatSlot := o.typ.slots.Float; floatSlot != nil {
+			result, raised := floatConvert(floatSlot, f, o)
+			if raised != nil {
+				return nil, raised
+			}
+			f := result.Value()
```
trotterdylan: I think you can ditch this temporary variable f and pass result.Value() directly into complex() below.

runtime/complex.go
```diff
@@ -130,3 +134,133 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+		}
+		s := toStrUnsafe(o).Value()
+		result, err := parseComplex(s)
+		if err != nil {
+			return nil, f.RaiseType(ValueErrorType, "complex() arg is a malformed string")
+		}
+		return NewComplex(result).ToObject(), nil
+	}
+
+	// TODO: fix me
```
trotterdylan: Flagging this TODO to make sure it gets addressed.

runtime/complex.go
```diff
@@ -130,3 +134,133 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+	fj := `(?P<imag3>` + fre + `)j`
+	f := `(?P<real4>` + fre + `)`
+	sj := `(?P<sign5>[-+])j`
+	j := `(?P<onlyJ>j)`
+	r := regexp.MustCompile(`^(?:` + fsfj + `|` + fsj + `|` + fj + `|` + f + `|` + sj + `|` + j + `)$`)
+	subs := r.FindStringSubmatch(ts)
+	if subs == nil {
+		return complex(0, 0), errors.New("Malformed complex string, no mathing pattern found")
+	}
+	const Real1 = 1
```
trotterdylan: Nitpick, declare these in a const (...) block. Also, don't capitalize them:

```
const (
    real1 = 1
    imag1 = 2
    ...
)
```

Also, probably a little easier to read this function if these constants are declared at the package scope, rather than in this function.

runtime/complex.go
```diff
@@ -130,3 +134,133 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+		return complex(0, 1), nil
+	}
+	if subs[OnlyJ] != "" {
+		return complex(0, 1), nil
+	}
+	return complex(0, 0), errors.New("Malformed complex string")
+}
+
+func unsignNaN(s string) string {
+	us := strings.ToUpper(s)
```
trotterdylan: Nitpick: In this codebase we usually use ToLower() for this kind of normalization.

m4ns0ur:
@trotterdylan I changed the structure based on [CPython Implementation](https://github.com/python/cpython/blob/18b250f844bf8b2d1a81c2d2dcc74e850364fe35/Objects/complexobject.c#L934) to avoid boilerplate
still believe it can be cleaned up, so please advice

runtime/complex_test.go
```diff
@@ -80,6 +83,109 @@ func TestComplexNE(t *testing.T) {
...
+		{args: wrapArgs(ComplexType, math.NaN()), want: NewComplex(complex(math.NaN(), 0)).ToObject()},
+		{args: wrapArgs(ComplexType, math.Inf(1)), want: NewComplex(complex(math.Inf(1), 0)).ToObject()},
+		{args: wrapArgs(ComplexType, math.Inf(-1)), want: NewComplex(complex(math.Inf(-1), 0)).ToObject()},
+		{args: wrapArgs(ComplexType, biggestFloat), want: NewComplex(complex(math.MaxFloat64, 0)).ToObject()},
+		{args: wrapArgs(ComplexType, new(big.Int).Neg(biggestFloat)), want: NewComplex(complex(-math.MaxFloat64, 0)).ToObject()},
+		{args: wrapArgs(ComplexType, new(big.Int).Sub(big.NewInt(-1), biggestFloat)), wantExc: mustCreateException(OverflowErrorType, "long int too large to convert to float")},
+		{args: wrapArgs(ComplexType, new(big.Int).Add(biggestFloat, big.NewInt(1))), wantExc: mustCreateException(OverflowErrorType, "long int too large to convert to float")},
+		{args: wrapArgs(ComplexType, bigLongNumber), wantExc: mustCreateException(OverflowErrorType, "long int too large to convert to float")},
+		// {args: wrapArgs(ComplexType, complex(1, 2)), want: NewComplex(complex(1, 2)).ToObject()},
```
m4ns0ur: @trotterdylan couldn't pass this test cases yet
something's fishy regarding to Complex Slot

runtime/complex_test.go
```diff
@@ -146,6 +252,61 @@ func TestComplexRepr(t *testing.T) {
...
+		{"((2.1-3.4j))", complex(0, 0), ErrSyntax},
+		{"3.14 -15.16j", complex(0, 0), ErrSyntax},
+		{"3.14- 15.16j", complex(0, 0), ErrSyntax},
+		{"3.14-15.16 j", complex(0, 0), ErrSyntax},
+		{"3.14 - 15.16 j", complex(0, 0), ErrSyntax},
+		{"foo", complex(0, 0), ErrSyntax},
+		{"foo+bar", complex(0, 0), ErrSyntax},
+	}
+	for _, cas := range cases {
+		if got, _ := parseComplex(cas.s); got != cas.want {
```
trotterdylan: I think you could use complexesAreSame() which was recently added:

```
if got, _ := parseComplex(cas.s); !complexesAreSame(got, cas.want) {
  t.Errorf(...)
}
```

runtime/complex.go
```diff
@@ -169,6 +178,185 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+			return nil, f.RaiseType(ValueErrorType, "complex() arg is a malformed string")
+		}
+		return NewComplex(result).ToObject(), nil
+	}
+	if argc == 2 && args[1].isInstance(StrType) {
+		return nil, f.RaiseType(TypeErrorType, "complex() second arg can't be a string")
+	}
+	var cr complex128
+	crIsComplex := false
+	if complexSlot := o.typ.slots.Complex; complexSlot != nil {
```
trotterdylan: Probably worth factoring this `if slots.Complex/else if slots.Float` piece out into a helper to eliminate some duplicate code.

runtime/complex.go
```diff
@@ -169,6 +178,185 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+	} else if floatSlot := args[1].typ.slots.Float; floatSlot != nil {
+		result, raised := floatConvert(floatSlot, f, args[1])
+		if raised != nil {
+			return nil, raised
+		}
+		ci = complex(result.Value(), 0)
+	} else {
+		return nil, f.RaiseType(TypeErrorType, "complex() argument must be a string or a number")
+	}
+	if ciIsComplex {
```
trotterdylan: Rather than doing all this bookkeeping, could you just do something like this:

```
cr, raised := complexConvert(args[0])
if raised != nil {
  return nil, raised
}
var ci complex128
if len(args) > 1 {
  if ci, raised = complexConvert(args[1]); raised != nil {
    return nil, raised
  }
}
return NewComplex(cr + ci * 1i).toObject(), nil
```

runtime/complex.go
```diff
@@ -169,6 +178,169 @@ func complexCoerce(o *Object) (complex128, bool) {
...
+		if raised != nil {
+			return complex(0, 0), raised
+		}
+		return complex(result.Value(), 0), nil
+	} else {
+		return complex(0, 0), f.RaiseType(TypeErrorType, "complex() argument must be a string or a number")
+	}
+}
+
+func complexConvert(complexSlot *unaryOpSlot, f *Frame, o *Object) (*Complex, *BaseException) {
```
trotterdylan: Please move helper functions to the end of the file. The general order of stuff in these files is: type definition, methods, type init function, helpers.

runtime/complex_test.go
```diff
@@ -132,6 +133,111 @@ func TestComplexNE(t *testing.T) {
...
+		{args: wrapArgs(ComplexType, "+17.59123j"), want: NewComplex(complex(0, 17.59123)).ToObject()},
+		{args: wrapArgs(ComplexType, "-3.0007e3j"), want: NewComplex(complex(0, -3000.7)).ToObject()},
+		{args: wrapArgs(ComplexType, "1+2j"), want: NewComplex(complex(1, 2)).ToObject()},
+		{args: wrapArgs(ComplexType, "3.1415-23j"), want: NewComplex(complex(3.1415, -23)).ToObject()},
+		{args: wrapArgs(ComplexType, "-23+3.1415j"), want: NewComplex(complex(-23, 3.1415)).ToObject()},
+		{args: wrapArgs(ComplexType, "+451.2192+384.27j"), want: NewComplex(complex(451.2192, 384.27)).ToObject()},
+		{args: wrapArgs(ComplexType, "-38.378-283.28j"), want: NewComplex(complex(-38.378, -283.28)).ToObject()},
+		{args: wrapArgs(ComplexType, "1.76123e2+0.000007e6j"), want: NewComplex(complex(176.123, 7)).ToObject()},
+		{args: wrapArgs(ComplexType, "-nan+nanj"), want: NewComplex(complex(math.NaN(), math.NaN())).ToObject()},
+		{args: wrapArgs(ComplexType, "inf-infj"), want: NewComplex(complex(math.Inf(1), math.Inf(-1))).ToObject()},
```
trotterdylan: Looks like this is the sole failing test. According to my testing in CPython, your expectations are correct so you might have to handle this edge case specially.

m4ns0ur: @trotterdylan not that one, but [this is the test case](https://github.com/google/grumpy/pull/296/files/0b650ff65a4b71dd0953e14a353aca01c3d070b2#diff-463afa6704d82d9d40baa2e71e0c1f63R206)
and here is the reason:
```
cr := complex(math.Inf(1), 0)
ci := complex(math.Inf(-1), 0)
fmt.Println(cr + ci*1i)
// getting (NaN-Infi), we expect (+Inf-Infi).
```
should I revert that correction part (just like [CPython](https://github.com/python/cpython/blob/master/Objects/complexobject.c#L1068))?

trotterdylan: That seems like strange behavior in Go complex arithmetic. I think the correct outcome can be had by using the expression complex(real(cr)-imag(ci), imag(cr)+real(ci))

trotterdylan: Also worth adding a comment about why we're using this expression instead of the simpler one we had originally.

m4ns0ur:
really happy that we have these _python compatibility_ tests 👍
- just implemented `complex.__neg__` for test cases like `assert complex('-j') == -1j`
- have to tune `repr()` for test cases like `assert repr(complex('nan-nanj')) == "(nan+nanj)"`, we already get `"(NaNNaNj)"` for this

trotterdylan:
@m4ns0ur: about the "strange assertion", https://github.com/google/grumpy/issues/262 may be relevant

m4ns0ur:
@trotterdylan yes, same issue here.
would you mind if I remove this assertion test case for now, to be resolved under #262 ?

trotterdylan:
Sounds good to me.

m4ns0ur:
@trotterdylan ready to review

trotterdylan:
Fantastic! Thanks for all your hard work on this. Merging.

PR #302, opened 2017-04-28
[WIP] Needed to fix #290: Allow sys.stdout to be replaceble

### LINKED ISSUE #290: 'print' statement ignores 'sys.stdout'
Replacing ``sys.stdout`` have no effect over ``print`` statement.

This test script outputs ``AssertionError: 0 chars printed, instead of 3``, after printing 'foo' in the stdout.

```python
import StringIO
import sys

sio = StringIO()
stdout = sys.stdout
sys.stdout = sio

print 'foo'

sys.stdout = stdout

chars = sio.tell()
assert chars == 3, '%s chars printed, instead of 3' % chars
```

alanjds:
Please refer to [the discussion](https://github.com/google/grumpy/issues/290#issuecomment-297933648) on #290

Do not merge until `__getattr__` and `__setattr__` is working.


corona10:
I have one thing that I want to discuss here. 
At this point, I think that we should create `module` directory for these kinds of task? `runtime/sysmodule.go` => `module/sysmodule.go`
WDYT? @trotterdylan @alanjds

alanjds:
I totally agree, but am not mature enough on Golang to glance the implications of this change.

corona10:
I believe that more modules will be needed. e.g (socketModule etc..)

alanjds:
Will continue the #290 implementation over this code.

Please discuss what changes we want before this (#302) got be merged.

trotterdylan:
I would avoid \_\_getattribute\_\_ and instead use \_\_getattr\_\_. The latter is called only after nothing is found through normal attribute lookup, which is probably what we want in this case.

alanjds:
Me too. But this rabbit hole is getting too deep!
Let me close some Print() issues first and later I come back to implement `__getattr__`.

This `__getattr__` issue seems at least one magnitude harder than the other issues anyway. I will need to get "slots" understanding first.

alanjds:
Just to say, tracebacks would be very nice. Took me quite a bit to realize this 'res' typo

alanjds:
> I would avoid __getattribute__ and instead use __getattr__. The latter is called only after nothing is found through normal attribute lookup, which is probably what we want in this case.

@trotterdylan Are you ok with going on this way, and allow to use __getattribute__ for now, or is better to go develop __getattr__ before keep going on this PR?

trotterdylan:
I think that getattr should be fairly straightforward to implement. It needs a new slot and fallback logic in grumpy.GetAttr(). Also a few tests that validate the correct attribute resolution.

alanjds:
Ok. Then I will chase \_\_getattr__ on another PR.

Everything else is ok?

trotterdylan:
I think this is a little more complicated than I had hoped it would be. What if we get rid of the hybrid module concept and instead swap out sys.\_\_dict\_\_ something like this:

```
# lib/sys.py
from __go__.grumpy import SysDict, SysModules
SysModules['sys'].__dict__ = SysDict
... carry on with the module as normal
```

Currently \_\_dict\_\_ is not settable, but it will be once https://github.com/google/grumpy/pull/331 is merged.

alanjds:
Hum... the whole point of this PR is allow ``sys.stdout = StringIO()`` to propagate to ``print``.

If ``SysModules['sys'].__dict__ = SysDict`` fixes this need and is simpler, lets do it!

Seems to still need an ``[SysDict[k] = locals()[k] for k in __all__]`` at the last line, but for me it looks more ok than the HybridModule stuff.

trotterdylan:
Right, so the trick would be get rid of the Stdout, Stdin and Stderr grumpy package members and instead pull those from SysDict like `SysDict.GetItemString(f, "stdin")`. I have a change started to see how this would work in practice. I can either finish that up and create a PR, or if you like, you can go that route in this PR.

> Seems to still need an [SysDict[k] = locals()[k] for k in __all__] at the last line, but for me it looks more ok than the HybridModule stuff.

I don't think this is necessary if we do the \_\_dict\_\_ assignment at the very top of the module.

PR #303, opened 2017-04-28
[WIP] Support for __getattr__ and __setattr__

alanjds:
Needed for #302.

And could allow cleaner generic integration of hybrid modules, composed of Golang+Python files.

### DIFF
--- a/testing/class_test.py
+++ b/testing/class_test.py
@@ -40,6 +40,8 @@ assert foo.bar() == 'bar'
 assert foo.baz() == 'bar'
 
 foo.b = 10
+assert foo.b == 10
+
 del foo.b
 assert not hasattr(foo, 'b')
 try:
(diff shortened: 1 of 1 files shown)


alanjds:
From the [CPython docs](https://docs.python.org/2/reference/datamodel.html#special-method-lookup-for-new-style-classes), it seems that some logic from grumpy ``object.__getattribute__`` should migrate to ``__builtins__.getattr``, specially the _implicit lookup_ part:

```python
>>> class Meta(type):
...    def __getattribute__(*args):
...       print "Metaclass getattribute invoked"
...       return type.__getattribute__(*args)
...
>>> class C(object):
...     __metaclass__ = Meta
...     def __len__(self):
...         return 10
...     def __getattribute__(*args):
...         print "Class getattribute invoked"
...         return object.__getattribute__(*args)
...
>>> c = C()
>>> c.__len__()                 # Explicit lookup via instance
Class getattribute invoked
10
>>> type(c).__len__(c)          # Explicit lookup via type
Metaclass getattribute invoked
10
>>> len(c)                      # Implicit lookup
10
```

alanjds:
@trotterdylan I guess this is good for now.

But I do think that getting Python tests count on coverage is so better on the long run that is worth to do this instead of working on Golang coverage to get this merged faster.

As said on [this comment](https://github.com/google/grumpy/pull/243#issuecomment-292216913), can you elaborate on how do you think this could be done?

Issue #282, opened 2017-04-06
TestRawInput() failed frequently.


corona10:
`TestRawInput` looks like have a significant issue.
Sometimes I can reproduce on my local environment.

```
--- FAIL: TestRawInput (0.00s)
	builtin_types_test.go:522: func1('HelloGrumpy\n', '') raised EOFError('EOF when reading a line',), want <nil>
Exception: foo
```

trotterdylan:
I worked on this for a little while but haven't yet figured out the problem. I was able to reproduce it more reliably if my machine was under load, e.g. by running tests in parallel.

There's two issues. The first is that these lines need to handle EOF explicitly:

```
	line, err := Stdin.reader.ReadString('\n')
	if err != nil {
		return nil, f.RaiseType(EOFErrorType, "EOF when reading a line")
	}
```

Should be something like:

```
	line, err := Stdin.reader.ReadString('\n')
	if err != nil && err != io.EOF {
		return nil, f.RaiseType(IOErrorType, err.Error())
	}
        if line == "" {
                return nil, f.RaiseType(EOFErrorType, "EOF when reading a line")
        }
```

When you do this you find that the error is actually "bad file". So Stdin is getting closed or something in the mean time. But I couldn't figure out why.

trotterdylan:
The test seems to be failing a lot more frequently on Travis lately, so I commented out the test in https://github.com/google/grumpy/commit/a76caef1a4e8172f2095380b82ee2d89eb1f3ede until it's fixed.

PR #304, opened 2017-05-01
Fixes #290: pyPrint accepts any filelike, with sys.stdout fallback

### LINKED ISSUE #290: 'print' statement ignores 'sys.stdout'

alanjds:
Gets sys.stdout from SysmoduleDict['stdout'], that exists and could be replaced (after #302 got merged)

pyPrint allows any filelike *Object to be printed to.

TBD: Factor out the method invocation logic to some ``func (o *Object) Invoke(f *Frame, method *Str)``

### DIFF
--- a/lib/sys.py
+++ b/lib/sys.py
@@ -88,10 +88,10 @@ class _SysModule(object):
             SysmoduleDict[k] = globals()[k]
     def __setattr__(self, name, value):
         SysmoduleDict[name] = value
-    def __getattr__(self, name):
+    def __getattribute__(self, name):   # TODO: replace w/ __getattr__ when implemented
         resp = SysmoduleDict.get(name)
         if res is None and name not in SysmoduleDict:
-            return super(_SysModule, self).__getattr__(name)
+            return super(_SysModule, self).__getattribute__(name)
         return resp
 
 modules = SysModules
--- a/lib/sys.py
+++ b/lib/sys.py
@@ -15,23 +15,36 @@
 """System-specific parameters and functions."""
 
 from __go__.os import Args
-from __go__.grumpy import SysModules, MaxInt, Stdin as stdin, Stdout as stdout, Stderr as stderr  # pylint: disable=g-multiple-import
+from __go__.grumpy import SysmoduleDict, SysModules, MaxInt, Stdin, Stdout, Stderr  # pylint: disable=g-multiple-import
 from __go__.runtime import Version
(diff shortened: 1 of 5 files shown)


alanjds:
Have no idea why this is breaking. Halp.
screen shot 2017-06-13 at 11 34 43

trotterdylan:
The types_test failure is because types.py defines ModuleType as type(sys) and types_test checks that it's the same type as __go__.grumpy.ModuleType. sys is now a different kind of object.

I would remove that line from types.py and add `from __go__.grumpy import ModuleType` to the top of that file.

The other issue is a transient failure due to a race condition in TestRunMain. I can trigger it reasonably consistently as follows:

```
$ GOPATH=$PWD/build go test grumpy -test.run=TestRunMain -count=100
--- FAIL: TestRunMain (0.00s)
	module_test.go:300: runMainRedirectStderr() failed: read |0: bad file descriptor
	module_test.go:300: runMainRedirectStderr() failed: read |0: bad file descriptor
	module_test.go:300: runMainRedirectStderr() failed: read |0: bad file descriptor
	module_test.go:300: runMainRedirectStderr() failed: read |0: bad file descriptor
FAIL
FAIL	grumpy	0.047s
```

I think this is similar to https://github.com/google/grumpy/issues/282. This is not due to your change. Just restart the build if you encounter that again.

alanjds:
OSX: Thanks. Will take a look on ``types.py``. I had used the ``import errno; type(errno)`` on the "hack module" as ``errno`` is a so simpler module. Is this acceptable for you?

Linux: Good to know. However, how can I restart the build on Travis? Last time I had sent a pointless commit to trigger it, but this adds noise to the history.

alanjds:
I had hit another strange error here, on itertools_test.py:
```
NameError: name '__name__' is not defined
exit status 1
```
Bisected and found its root on 35fc2a3, but it makes no sense.

As this PR will not be merged until \_\_getattr__ is working, will leave it as is. This seems as a symptom of another undiscovered bug on the module/types stuff.

PR #309, opened 2017-05-13
Implement complex.__abs__

m4ns0ur:


testing/builtin_test.py
```diff
@@ -28,6 +28,19 @@
 assert abs(-3.4) == 3.4
 assert isinstance(abs(-3.4), float)
 
+assert abs(complex(0, 0)) == 0.0
```
trotterdylan: At least one of these assertions is triggering. I verified that they all pass in CPython so it's likely something Go specific.

m4ns0ur: just removed 2 of assertions which are related to #262, will create a PR soon on #262 to solve it once forever

runtime/float.go
```diff
@@ -260,7 +261,8 @@ func floatRDivMod(f *Frame, v, w *Object) (*Object, *BaseException) {
 }
 
 func floatRepr(f *Frame, o *Object) (*Object, *BaseException) {
-	return NewStr(strconv.FormatFloat(toFloatUnsafe(o).Value(), 'g', -1, 64)).ToObject(), nil
+	v := unsignPositiveInf(strings.ToLower(strconv.FormatFloat(toFloatUnsafe(o).Value(), 'f', -1, 64)))
+	return NewStr(v).ToObject(), nil
```
trotterdylan: This causes other problems :(

From Travis: AssertionError: u'1e+06' != '1000000'

m4ns0ur:
The question is, why `u'1e+06'` is falling back in `float.__repr__`? I should check unicode implementation

runtime/float.go
m4ns0ur: The question is, why u'1e+06' is falling back in float.__repr__? I should check unicode implementation

m4ns0ur:
@trotterdylan just fixed the `float.__repr__`, but we still need to modify some of assertions in `expr_visitor_test.py`, to be like this:
```
testNumFloatNoDecimal = _MakeLiteralTest('5.', '5.0')
testNumFloatSci = _MakeLiteralTest('1e16', '1e+16')
testNumFloatSciCap = _MakeLiteralTest('1E16', '1e+16')
testNumFloatSciCapPlus = _MakeLiteralTest('1E+16', '1e+16')
```
from CPython:
```
$ python -c "print repr(5.)"
5.0
$ python -c "print repr(1e+06)"
1000000.0
$ python -c "print repr(1e+15)"
1000000000000000.0
$ python -c "print repr(1e+16)"
1e+16
```

runtime/float.go
```diff
@@ -259,8 +261,14 @@ func floatRDivMod(f *Frame, v, w *Object) (*Object, *BaseException) {
 	})
 }
 
+const floatReprPrecision = 16
+
 func floatRepr(f *Frame, o *Object) (*Object, *BaseException) {
-	return NewStr(strconv.FormatFloat(toFloatUnsafe(o).Value(), 'g', -1, 64)).ToObject(), nil
+	v := unsignPositiveInf(strings.ToLower(strconv.FormatFloat(toFloatUnsafe(o).Value(), 'g', floatReprPrecision, 64)))
+	if noDecimal, _ := regexp.MatchString(`^\d+$`, v); noDecimal {
```
trotterdylan: This check seems pretty heavyweight for what we need to do here. At the very least we should compile the expression so that we don't do that in every call. But I think avoiding the regexp dependency altogether by using strings.IndexFunc() or something would be cleaner.

m4ns0ur: welcome back! just moved to its own CR

m4ns0ur:

trotterdylan:
Looks great! Thanks for the PR. Merging.

PR #319, opened 2017-06-05
Implement float.__str__ (also fix #262)

### LINKED ISSUE #262: float.__str__ should return only 12 digits of precision

m4ns0ur:
It will fix #262 as well


corona10:
Awesome!

PR #342, opened 2017-07-01
Add user module

m4ns0ur:


trotterdylan:
Given that execfile isn't/won't be supported by Grumpy and this module is removed from Python 3 anyway, I wonder if we should just drop this module in the Grumpy stdlib.

PR #348, opened 2017-07-04
Return only status on closing popen()

m4ns0ur:
It should return only `status`, on closing the `popen()`:
```
>> python -c "import os; print os.popen('foo 2>&1').close()"
32512

>> echo "import os; print os.popen('foo 2>&1').close()"|make run
(23105, -1)
```

### DIFF
--- a/lib/os/__init__.py
+++ b/lib/os/__init__.py
@@ -136,10 +136,7 @@ class _Popen(object):
     state, err = self.result
     if err:
       raise OSError(err.Error())
-    result = _encode_wait_result(state.Sys())
-    if not result:
-      return None
-    return result 
+    return state.Sys() 
 
 
 def popen(command, mode='r'):
--- a/lib/os/__init__.py
+++ b/lib/os/__init__.py
@@ -139,7 +139,7 @@ class _Popen(object):
     result = _encode_wait_result(state.Sys())
     if not result:
       return None
-    return state.status 
+    return result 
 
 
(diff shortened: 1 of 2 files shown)


lib/os/__init__.py
```diff
@@ -139,7 +139,7 @@ def close(self, _):
     result = _encode_wait_result(state.Sys())
     if not result:
       return None
-    return state.Pid(), result
+    return state.status
```
trotterdylan: I think you should return `result` here, which should be the same as state.status on most platforms.

m4ns0ur: well, my Debian machine is not working, and I was testing on a Windows 10 machine (Bash/WSL) :-( and result is always `-1`.

m4ns0ur: Ok, I should update the test, since result would be `None` on successful.
I'm not sure it's a good idea to add a test case for a fail condition (e.g. `command not found` with result `32512`)? Is it going to be equal on different platforms?

trotterdylan: Hmm looking at this a bit more it seems like state.Sys() (which is the same thing as state.status) is encoded the way we want (exit status in high byte and signal in low byte) on OS X so we could get rid of the _encode_wait_result() call and just use state.Sys() (which is already a number) directly.

Most commands will not be consistent across platforms for testing purposes, but I think the result of `exit N` for whatever N should work on most reasonable platforms for testing purposes.

PR #351, opened 2017-07-07
Implement string.rjust() ljust() and center().

corona10:
* Support center(), ljust() and rjust().
* TypeError messages are following CPython 2.7.12
* Add unit tests.


corona10:
@trotterdylan PTAL

runtime/str.go
```diff
@@ -699,6 +741,26 @@ func strRIndex(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+
+	if numChars := len(fill); numChars != 1 {
+		return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("rjust() argument 2 must be char, not str"))
+	}
+
+	if len(s) >= width {
+		return NewStr(s).ToObject(), nil
+	}
+
+	return NewStr(pad(s, width-len(s), 0, fill)).ToObject(), nil
```
trotterdylan: Seems like everything above this line is common to all the methods except for the name. Maybe factor it out into strJustDecodeArgs or something

runtime/str_test.go
```diff
@@ -306,6 +306,11 @@ func TestStrMethods(t *testing.T) {
 		{"capitalize", wrapArgs("вол"), NewStr("вол").ToObject(), nil},
 		{"capitalize", wrapArgs("foobar", 123), nil, mustCreateException(TypeErrorType, "'capitalize' of 'str' requires 1 arguments")},
 		{"capitalize", wrapArgs("ВОЛ"), NewStr("ВОЛ").ToObject(), nil},
+		{"center", wrapArgs("foobar", 10, "#"), NewStr("##foobar##").ToObject(), nil},
```
trotterdylan: How about a test case where the number of characters to pad are odd. Which side should the extra character go on?

corona10:
@trotterdylan Thanks for the review. Updated!

PR #353, opened 2017-07-08
Implement file.closed()

m4ns0ur:


runtime/file.go
```diff
@@ -217,6 +217,16 @@ func fileClose(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
 	return ret, nil
 }
 
+func fileClosed(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
+	if raised := checkMethodArgs(f, "closed", args, FileType); raised != nil {
+		return nil, raised
+	}
+	file := toFileUnsafe(args[0])
+	file.mutex.Lock()
+	defer file.mutex.Unlock()
```
trotterdylan: Avoid defer for simple cases like this because it's very inefficient. In this case just store the result in a variable, unlock, then return the variable.

m4ns0ur:
it's ready to review

PR #354, opened 2017-07-08
Return bool on stat.S_ISDIR()

m4ns0ur:
`stat.S_ISDIR()` should return `boolean`:
```
>> python -c "import stat; print stat.S_ISDIR(511)"
False

>> echo "import stat; print stat.S_ISDIR(511)"|make run
0
```

### DIFF
--- a/third_party/stdlib/test/test_stat.py
+++ b/third_party/stdlib/test/test_stat.py
@@ -82,92 +82,97 @@ class TestFilemode(unittest.TestCase):
             else:
                 self.assertFalse(func(mode))
 
-#   def test_mode(self):
-#       with open(TESTFN, 'w'):
-#           pass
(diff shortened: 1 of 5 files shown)


third_party/stdlib/test/test_stat.py
```diff
@@ -0,0 +1,175 @@
...
+            if func is None:
+                if funcname == testname:
+                    raise ValueError(funcname)
+                continue
+            if funcname == testname:
+                self.assertTrue(func(mode))
+            else:
+                self.assertFalse(func(mode))
+
+#   def test_mode(self):
```
corona10: Please mark `@unittest.skip('grumpy')` for commented out tests.
Maybe @trotterdylan want this. :)

PR #360, opened 2017-07-12
Add test for contextlib module

m4ns0ur:


third_party/stdlib/test/test_contextlib.py
```diff
@@ -0,0 +1,353 @@
...
+            state.append(1)
+            yield 42
+            state.append(999)
+        with woohoo() as x:
+            self.assertEqual(state, [1])
+            self.assertEqual(x, 42)
+            state.append(x)
+        self.assertEqual(state, [1, 42, 999])
+
+    @unittest.skip('grumpy')
```
trotterdylan: ISTM that most of these tests should work. Is there something specific that's broken preventing them from working?

m4ns0ur: Yes, 'stat._getframe()' is missing.
not sure how deap it's gonna be, but I can give it a shot.

trotterdylan: Note that Grumpy has a built in called \_\_frame\_\_() that gets the top level fame and you can use f_back to get subsequent ones.

m4ns0ur:
@trotterdylan it looks better now, PTAL.

BTW couple of test cases are failing, because raising an error in `with - generator` block, is not triggering `generator - finally`, any thought why?

runtime/dict_test.go
```diff
@@ -615,6 +615,19 @@ func TestDictNewRaises(t *testing.T) {
 	}
 }
 
+func TestDictSetDefault(t *testing.T) {
+	cases := []invokeTestCase{
+		{args: wrapArgs(newTestDict("foo", 42), "foo"), want: NewInt(42).ToObject()},
```
trotterdylan: You should also check that the dictionary is updated appropriately. One way to do that is to define a function that returns the result of setdefault and the dictionary itself and comparing against that. See TestDictSetItem for an example.

m4ns0ur:
it's too far from master already, so I'm closing this one. will create new one after `generator.throw()` PR.

PR #362, opened 2017-07-16
Implement sys._getframe

m4ns0ur:


lib/sys_test.py
```diff
@@ -86,6 +86,15 @@ def TestExitInvalidArgs():
     assert False
 
 
+def test_getframe(self):
+  self.assertRaises(TypeError, sys._getframe, 42, 42)
+  self.assertRaises(ValueError, sys._getframe, 2000000000)
+  self.assertTrue(
+    SysModuleTest.test_getframe.im_func.func_code \
```
trotterdylan: I don't think this test is running because this file uses the weetest library for unit testing, not the standard unittest. This is a holdover from before unittest was supported.

Basically, the function should be named like TestGetFrame() and there is no self or assert* methods. Just use assert statements.

m4ns0ur: @trotterdylan done! PTAL

lib/sys_test.py
```diff
@@ -86,13 +86,17 @@ def TestExitInvalidArgs():
...
-  self.assertRaises(ValueError, sys._getframe, 2000000000)
-  self.assertTrue(
-    SysModuleTest.test_getframe.im_func.func_code \
-    is sys._getframe().f_code
-  )
+def TestGetFrame():
+  try:
+    sys._getframe(42, 42)
+  except TypeError:
+    assert True
```
trotterdylan: I would make this a pass statement and put an else block with assert False so that if no exception is raised the assertion triggers. Ditto below.

m4ns0ur: @trotterdylan PTAL

PR #365, opened 2017-07-17
Implement dict.setdefault

m4ns0ur:


runtime/dict.go
```diff
@@ -755,6 +755,35 @@ func dictRepr(f *Frame, o *Object) (*Object, *BaseException) {
...
+	expectedTypes := []*Type{DictType, ObjectType, ObjectType}
+	if argc == 2 {
+		expectedTypes = expectedTypes[:2]
+	}
+	if raised := checkMethodArgs(f, "setdefault", args, expectedTypes...); raised != nil {
+		return nil, raised
+	}
+	k := args[1]
+	d := toDictUnsafe(args[0])
+	i, raised := d.GetItem(f, k)
```
trotterdylan: Doing a get and set leaves us open to a race condition if the dict changes in the mean time. I think we need to lock over the whole operation. I suspect it will be quite similar to putItem so some refactoring may be in order.

Incidentally that may allow us to pass the atomic test you had to skip.

m4ns0ur:
@trotterdylan PTAL

runtime/dict.go
```diff
@@ -755,6 +747,54 @@ func dictRepr(f *Frame, o *Object) (*Object, *BaseException) {
...
+			// of trying to recover.
+			raised = f.RaiseType(RuntimeErrorType, "dictionary changed during write")
+		} else {
+			if entry == nil {
+				if argc > 2 {
+					value = args[2]
+				} else {
+					value = None
+				}
+				raised = addDictTableEntry(f, d, hash, key, value, index)
```
trotterdylan: Factoring this out is a good first step. I'm wondering though if we can go even further: do you think it's possible for putItem to take a function that will replace the table manipulation from line 409-426? I'm thinking that the callers that are deleting a key can pass a function that does this part:

```
if entry != nil && entry != deletedEntry {
	d.table.storeEntry(index, deletedEntry)
	d.table.incUsed(-1)
	d.incVersion()
}
```

While the callers that are setting a key can pass a function that will do this part:

```
newEntry := &dictEntry{hash.Value(), key, value}
if newTable, ok := t.writeEntry(f, index, newEntry); ok {
	if newTable != nil {
		d.storeTable(newTable)
	}
	d.incVersion()
} else {
	raised = f.RaiseType(OverflowErrorType, errResultTooLarge)
}
```

Finally, we could have a slightly different function for setdefault, something like:

```
if entry == nil || entry == deletedEntry {
	return
}
newEntry := &dictEntry{hash.Value(), key, value}
if newTable, ok := t.writeEntry(f, index, newEntry); ok {
	if newTable != nil {
		d.storeTable(newTable)
	}
	d.incVersion()
} else {
	raised = f.RaiseType(OverflowErrorType, errResultTooLarge)
}
```

The nice thing is that it encapsulates the locking, hashing and finding the entry. The function being called can then do its work in a safe, atomic way.

What do you think?

m4ns0ur: @trotterdylan sounds great!

m4ns0ur: @trotterdylan ready to review

runtime/dict.go
```diff
@@ -390,7 +390,8 @@ func (d *Dict) Len() int {
 
 // putItem associates value with key in d, returning the old associated value if
 // the key was added, or nil if it was not already present in d.
-func (d *Dict) putItem(f *Frame, key, value *Object) (*Object, *BaseException) {
+// fun is defining the operation on d, which can be add or delete the entry.
+func (d *Dict) putItem(f *Frame, key, value *Object, fun func(f *Frame, d *Dict, entry *dictEntry, hash *Int, key, value *Object, index int) *BaseException) (*Object, *BaseException) {
```
trotterdylan: Hmm this didn't work out quite the way I hoped: the callback is pretty inelegant because of all the parameters it requires. Sorry to run around in circles on this, but I think we should find another way.

What do you think about passing a new `overwrite bool` parameter to putItem() that, when false, will do the setdefault behavior, i.e. will only write the value if no value is already present for that key.

Since we already return the original value, we can return it from setdefault.

m4ns0ur: yes, I didn't really like that neither, that's a lot of parameters.
totally agree with `overwrite` parameter, at least we can make setdefault() cleaner, and it's not gonna mess up putitem() anyway.
let me revert this.

m4ns0ur:
@trotterdylan PTAL

runtime/dict.go
```diff
@@ -407,22 +407,9 @@ func (d *Dict) putItem(f *Frame, key, value *Object) (*Object, *BaseException) {
...
-				newEntry := &dictEntry{hash.Value(), key, value}
-				if newTable, ok := t.writeEntry(f, index, newEntry); ok {
-					if newTable != nil {
-						d.storeTable(newTable)
-					}
-					d.incVersion()
-				} else {
-					raised = f.RaiseType(OverflowErrorType, errResultTooLarge)
-				}
+				deleteDictEntry(d, entry, index)
```
trotterdylan: Would you mind also moving the code back into putItem()? Generally I prefer to having code inline rather than adding helpers that are only called in one place.

m4ns0ur: makes sense

m4ns0ur: done!

m4ns0ur:

PR #374, opened 2017-08-01
Add builtin filter function

aisk:


runtime/builtin_types.go
```diff
@@ -248,6 +249,82 @@ func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		}
+		var result bytes.Buffer
+		for _, item := range toStrUnsafe(l).Value() {
+			ret, raised := fn.Call(f, []*Object{NewStr(string(item)).ToObject()}, nil)
+			if raised != nil {
+				return nil, raised
+			}
+			if ret, raised := IsTrue(f, ret); raised != nil {
+				return nil, raised
+			} else if ret {
```
trotterdylan: Remove the else

runtime/builtin_types.go
```diff
@@ -248,6 +249,82 @@ func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+				}
+				if ret, raised := IsTrue(f, ret); raised != nil {
+					return nil, raised
+				} else if ret {
+					result = append(result, item)
+				}
+			}
+		}
+		return NewTuple(result...).ToObject(), nil
+	case l.isInstance(StrType):
```
trotterdylan: Does this behavior apply to unicode as well? Should this be basestring?

runtime/builtin_types.go
```diff
@@ -248,6 +249,82 @@ func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+		if fn == None {
+			return l, nil
+		}
+		var result bytes.Buffer
+		for _, item := range toStrUnsafe(l).Value() {
```
trotterdylan: This will iterate over l assuming it's a UTF-8 encoded string. I think what you actually want is to convert this to a []byte and iterate over each byte and WriteByte() below instead of WriteRune().

runtime/builtin_types.go
```diff
@@ -248,6 +249,82 @@ func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+			}
+		}
+		return NewStr(result.String()).ToObject(), nil
+	default:
+		result := make([]*Object, 0)
+		raised := seqForEach(f, l, func(item *Object) (raised *BaseException) {
+			if fn == None {
+				if ret, raised := IsTrue(f, item); raised != nil {
+					return raised
+				} else if ret {
```
trotterdylan: remove else

runtime/builtin_types.go
```diff
@@ -248,6 +249,82 @@ func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+					result = append(result, item)
+				}
+			} else {
+				ret, raised := fn.Call(f, []*Object{item}, nil)
+				if raised != nil {
+					return raised
+				}
+				if ret, raised := IsTrue(f, ret); raised != nil {
+					return raised
+				} else if ret {
```
trotterdylan: remove else

runtime/builtin_types.go
```diff
@@ -248,6 +249,82 @@ func builtinMapFn(f *Frame, args Args, _ KWArgs) (*Object, *BaseException) {
...
+				return nil, raised
+			} else if ret {
+				result.WriteRune(item)
+			}
+		}
+		return NewStr(result.String()).ToObject(), nil
+	default:
+		result := make([]*Object, 0)
+		raised := seqForEach(f, l, func(item *Object) (raised *BaseException) {
+			if fn == None {
```
trotterdylan: Given you do this check a few times I wonder if you should do something like this at the top:

```
filterFunc := IsTrue
if fn != None {
	filterFunc = func(f *Frame, o *Object) (bool, *BaseException) {
		result, raised := fn.Call(f, Args{o}, nil)
		if raised != nil {
			return false, raised
		}
		return IsTrue(f, result)
	}
}
```

And then use filterFunc throughout.

runtime/builtin_types_test.go
```diff
@@ -186,6 +186,11 @@ func TestBuiltinFuncs(t *testing.T) {
 		{f: "divmod", args: wrapArgs(-3.25, -1.0), want: NewTuple2(NewFloat(3.0).ToObject(), NewFloat(-0.25).ToObject()).ToObject()},
 		{f: "divmod", args: wrapArgs(NewStr("a"), NewStr("b")), wantExc: mustCreateException(TypeErrorType, "unsupported operand type(s) for divmod(): 'str' and 'str'")},
 		{f: "divmod", args: wrapArgs(), wantExc: mustCreateException(TypeErrorType, "'divmod' requires 2 arguments")},
+		{f: "filter", args: wrapArgs(None, NewTuple2(NewInt(0).ToObject(), NewInt(1).ToObject())), want: NewTuple1(NewInt(1).ToObject()).ToObject()},
+		{f: "filter", args: wrapArgs(BoolType, NewTuple2(NewInt(0).ToObject(), NewInt(1).ToObject())), want: NewTuple1(NewInt(1).ToObject()).ToObject()},
+		{f: "filter", args: wrapArgs(IntType, "012"), want: NewStr("12").ToObject()},
+		{f: "filter", args: wrapArgs(None, newTestList(1, 0, 3)), want: newTestList(1, 3).ToObject()},
+		{f: "filter", args: wrapArgs(IntType, newTestList("1", "0", "3")), want: newTestList("1", "3").ToObject()},
```
trotterdylan: Can you add tests for subclasses of str and tuple that implement \_\_iter\_\_ to make sure we maintain compatibility with CPython?

aisk: Hi can i add these tests in `testing/` directory using Python?

trotterdylan: I'd like tests in the Go code because it's easy to measure coverage there. Additional tests in testing/ are fine though.

aisk:
@trotterdylan Updated!

PR #379, opened 2017-09-12
Implement list.__delslice__

m4ns0ur:


runtime/list.go
```diff
@@ -228,6 +228,38 @@ func listDelItem(f *Frame, o *Object, key *Object) *BaseException {
...
+	if argc != 3 {
+		return nil, f.RaiseType(TypeErrorType, fmt.Sprintf("function takes exactly 2 arguments (%d given)", argc-1))
+	}
+	l := toListUnsafe(args[0])
+	if args[1].typ.slots.Index == nil || args[2].typ.slots.Index == nil {
+		return nil, f.RaiseType(TypeErrorType, "an integer is required")
+	}
+	l.mutex.Lock()
+	numListElems := len(l.elems)
+	start, raised := calcIndex(f, args[1], numListElems)
```
trotterdylan: Can we instead construct a slice and call DelSlice?

m4ns0ur: @trotterdylan was thinking same, but some tests are not passing:
```
--- FAIL: TestListDelSlice (0.00s)
	list_test.go:122: TestListDelSlice([1, 2, 3, 4, 5], -1, 3) = [1, 2, 3, 4, 5], want [4, 5]
	list_test.go:122: TestListDelSlice([1, 2, 3, 4, 5], 1, -3) = [1, 3, 4, 5], want [1, 2, 3, 4, 5]
```
from CPython:
```
>>> x=[1,2,3,4,5]
>>> x.__delslice__(-1, 3)
>>> x
[4, 5]
>>> x=[1,2,3,4,5]
>>> x.__delslice__(1, -3)
>>> x
[1, 2, 3, 4, 5]
>>> 
```

m4ns0ur: @trotterdylan ok, I will keep index alignment before initiating `Slice`, it should be enough

m4ns0ur:
@trotterdylan PTAL

PR #404, opened 2018-05-19
pip-installable package

alanjds:
Changes the folders structure separating the Tools and the Runtime+Stdlib. Allows to test and publish both parts separately. The Tools are tested via `pytest`, not the Makefile.

The Tools are to be published as `grumpy-tools` package and Runtime as `grumpy-runtime` package.

TODO:
- [X] Publish-able on Python Package Index (PyPI)
- [x] Update the README
- [x] Better way to bump the version then editing files. (Maybe via `bumpversion`?)
- [x] ~Fix a `find: 'compiler': No such file or directory` Makefile warning~


alanjds:
I gave up on `find: 'compiler': No such file or directory`. Too little `make` skills here; too advanced Makefile there.

Contributions on this front are appreciated ;)

I do consider this PR merge-able.

